PR #8, opened 2021-07-18
small revises on getting started

jackymao:
Here are the small revises on getting started tutorial:  

- fixed the link of "examples/getting_started" (so it shows correctly on the .github.io site)
- fixed typo "impl"
- clearly state codes position for beginner (like me, :): lib.rs, models.rs, toml binary, src/bin/*

That's it. Thanks for your attention.


Electron100 (approved):
Looks good, thanks for improving the documentation!

PR #10, opened 2021-08-01
Add offset parameter

DimmKG:
Offset allows you to skip a certain number of records before returning the result. This can be useful when requesting the next page of records if a limit is set.
 
Example
```rust
pub fn article_list(limit: i32, offset: i32, conn: &Connection) -> Result<Vec<ArticleOutput>, Error>
{
    let res = Query::<Article>::new("article")
        .limit(limit)
        .offset(offset)
        .load(conn);
    match res {
        Ok(articles) => {
            let mut result : Vec<ArticleOutput> = vec!();
            for article in articles {
                result.push(article.to_result(conn));
            }
            Ok(result)
        },
        Err(err) => return Err(err)
    }
}
```

### DIFF
--- a/butane_core/src/db/connmethods.rs
+++ b/butane_core/src/db/connmethods.rs
@@ -18,6 +18,7 @@ pub trait ConnectionMethods {
         columns: &'b [Column],
         expr: Option<BoolExpr>,
         limit: Option<i32>,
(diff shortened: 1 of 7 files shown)


Electron100 (approved):
Thank you for adding this!

PR #39, opened 2023-02-04
Tidy toml

jayvdb:
This is the result of running `cargo make format-toml` , which invokes https://github.com/sagiegurari/cargo-make , which invokes https://crates.io/crates/taplo-cli , which has a config file that can be used to alter the way toml files are formatting.

I like to enable `reorder_keys = true`, but that also alpha orders the `[package]` metadata section, and I've found that is an acquired taste.

I am happy to set up CI to enforce this, and even happier to set up a PR to replace `Makefile` with `Makefile.toml`, which has lots of Rust-specific advantages in addition to not using tabs.

### DIFF
--- a/butane/Cargo.toml
+++ b/butane/Cargo.toml
@@ -30,21 +30,21 @@ butane_core = { path = "../butane_core", version = "0.5" }
 
 [dev-dependencies]
 cfg-if = "1.0"
-exec_time = { version="0.1.4" }
+exec_time = { version = "0.1.4" }
 paste = "1.0.11"
-chrono = { version = "0.4", features=["serde"] }
+chrono = { version = "0.4", features = ["serde"] }
 env_logger = "0.10"
 geo-types = "0.7"
 libc = "0.2"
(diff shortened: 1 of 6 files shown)


Electron100:
I'd be fine with having the Taplo formatter check running in CI. 

I'll admit I have a harder time getting excited about cargo-make. Maybe I'm just a cranky traditionalist, and I'll certainly admit that GNU Make is getting long in the tooth in a lot of areas, but cargo-make's way of specifying the commands to run feels like a step backwards -- awkward compared to just writing the command the same way you would in a shell. 
The command invocation looks awkward to me too since as far as I can tell from a brief peruse through the docs there's no standard naming for the tasks file, so it has to be specified on the command-line each time?

jayvdb:
The default name is `Makefile.toml` - ya, I can see that the docs about that are confusing.  I'll raise that upstream.
The way to specify commands is quite verbose.  Anyway, I'll do a PR at some point, not right now, with no concerns if you reject it.

PR #40, opened 2023-02-04
Alpha sort keys in Cargo.toml files

jayvdb:
Followup to https://github.com/Electron100/butane/pull/39, this uses `reorder_keys`.  No worries if you reject because it also changes the package metadata sections.
If rejected, I'll see if I can find another tool which can reorder only the dependencies.

### DIFF
--- a/.taplo.toml
+++ b/.taplo.toml
@@ -1,2 +1,8 @@
 [formatting]
 reorder_arrays = true
+reorder_keys = true
+
+# Do not reorder keys within `package or `bin`.
+[[rule]]
+formatting = { reorder_keys = false }
+keys = ["bin", "package"]
--- a/butane_codegen/Cargo.toml
+++ b/butane_codegen/Cargo.toml
@@ -12,10 +12,10 @@ repository = "https://github.com/Electron100/butane"
 datetime = []
 
 [dependencies]
-proc-macro2 = { version = "1.0", default-features = false }
 butane_core = { path = "../butane_core", version = "0.5" }
+proc-macro2 = { version = "1.0", default-features = false }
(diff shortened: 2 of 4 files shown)


Electron100:
Hmm, I'm going to have to think about this one a little bit. The ordering of dependencies is nice (I've been trying to keep them ordered manually, but obviously I haven't fully succeeded), but I don't love the re-ordering of other sections.

butane/Cargo.toml
```diff
@@ -1,26 +1,26 @@
...
 
 [features]
-default = ["datetime", "json", "uuid"]
-json = ["butane_core/json"]
-sqlite = ["butane_core/sqlite"]
-sqlite-bundled = ["butane_core/sqlite-bundled"]
-pg = ["butane_core/pg"]
 datetime = ["butane_core/datetime", "butane_codegen/datetime"]
 debug = ["butane_core/debug"]
+default = ["datetime", "json", "uuid"]
```
jayvdb, line 17: putting this in the middle is also a significant degradation IMO.  I suspect most Rust developers look at Cargo.toml of libraries they are using, and the default feature list is very commonly what they are looking for.

I've created a bunch of issues like this https://github.com/tbrand/cargo-tomlfmt/issues/25 , as that tool is more likely to support Cargo.toml specific formatting than the more general `taplo`

jayvdb: More tools at https://github.com/TimonPost/cargo-toml-format#similar-work .  I'll try them out and see what works best.

jayvdb:
I've updated this.  I havent built CI to enforce it yet - there are only a few contributors atm, so not a high priority atm.

Electron100 (approved):

Issue #32, opened 2023-01-23
HashMap support


jayvdb:
HashMap's are not supported yet.
In postgres they can be implemented as a `hstore`.
In sqlite, they could be a separate table, e.g. roughly like `Many`, or serialised to a single field using JSON, i.e. https://github.com/Electron100/butane/issues/27

Electron100:
Hmm, I'm a little torn on this one. I'm sure the folks at Postgres had good reasons for adding the `hstore` type but on the other hand Postgres has a _lot_ of non-standard SQL types and I'm a bit reluctant to support them all out of the box.

I think ideally what I'd like to do here is improve support for custom types and allow `ToSql`/`FromSql` to provide different types based on the backend. That would make it possible to write a SQL transformations for things like this that used Postgres's extra types for that backend but didn't push a new core type into other backends that they needed to implement. I haven't fleshed out the details there though I'll admit.

I'm not necessarily opposed to this, but I would be interested to hear a little bit more about your use case if it's something you're able to talk about or summarize publicly -- when do you store HashMaps in a database?

jayvdb:
A HashMap is one of the OpenAPI/JSONSchema container types. c.f. https://stackoverflow.com/questions/40750340/how-to-define-json-schema-for-mapstring-integer  It is supported by https://github.com/oxidecomputer/progenitor and https://github.com/juhaku/utoipa .  It would be nice to be able to store these objects in the database.

All backends could implement the HashMap storage using JSON.  That seems to be the approach taken at https://github.com/SeaQL/sea-orm/issues/279 and https://github.com/diesel-rs/diesel/issues/970 .  Always using JSON will avoid any limitations of `hstore`.

Electron100:
Hmm, part of what I'm asking is: does the underlying Postgres hstore type matter to you? Do you want to be able to write SQL queries that depends on that underlying structure? Or do you just want to read/write a HashMap from/to a database as-is and the representation doesn't matter: could be json, could be a blob, or whatever?

If the latter then I think we can just do a straight ToSql/FromSql impl without regard to the backend.

jayvdb:
I dont mind what the underlying data storage is. Happy to be avoiding hstore to be honest.

jayvdb:
With https://github.com/Electron100/butane/pull/43 , the following badly written test passes.

```rust
#[butane_type(Json)]
#[derive(PartialEq, Eq, Debug, Clone)]
struct StringMap(HashMap<String, String>);

impl StringMap {
[...]

PR #45, opened 2023-02-05
Support HashMap as implicit type

jayvdb:
Closes https://github.com/Electron100/butane/issues/32

### DIFF
--- a/butane_core/src/migrations/adb.rs
+++ b/butane_core/src/migrations/adb.rs
@@ -113,6 +113,12 @@ impl TypeResolver {
         }
     }
     fn find_type(&self, key: &TypeKey) -> Option<TypeIdentifier> {
+        #[cfg(feature = "json")]
+        if let TypeKey::CustomType(ct) = key {
+            if ct.starts_with("HashMap < String,") {
+                return Some(TypeIdentifier::from(SqlType::Json));
+            }
+        }
         self.types.get(key).cloned()
     }
     fn insert(&mut self, key: TypeKey, ty: TypeIdentifier) -> bool {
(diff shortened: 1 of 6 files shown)


butane_core/src/sqlval.rs
```diff
@@ -472,6 +474,44 @@ impl FieldType for serde_json::Value {
 #[cfg(feature = "json")]
 impl PrimaryKeyType for serde_json::Value {}
 
+#[cfg(feature = "json")]
+impl ToSql for HashMap<String, String> {
```
jayvdb, line 478: This will hopefully become
```rust
impl<T> FromSql for HashMap<String, T>
where
    T: Clone + PartialEq,
```

butane_core/src/migrations/adb.rs
```diff
@@ -113,6 +113,12 @@ impl TypeResolver {
         }
     }
     fn find_type(&self, key: &TypeKey) -> Option<TypeIdentifier> {
+        #[cfg(feature = "json")]
+        if let TypeKey::CustomType(ct) = key {
+            if ct.starts_with("HashMap < String,") {
```
jayvdb, line 118: This is a horrible hack to avoid needing `butane_type`, which would require a `newtype` for each type of HashMap.

Electron100: Yeah, this is another great example of the limitations of trying to create the schema representation at compile time with only the AST and therefore why I'd like to move it to runtime.

Electron100: Since I haven't done that yet though, I'm reluctantly OK with the hack for now

Issue #44, opened 2023-02-05
Old types and tables left in .butane migrations directory


jayvdb:
When I commented out "StringMap" from my playground for https://github.com/Electron100/butane/issues/32#issuecomment-1417087644 , and I run the tests again, `butane/.butane/migrations/current/types.json` still has the entry from `CT:StringMap`.

Likewise if I add a new `#[model]` to the tests, and then comment it out, the `butane/.butane/migrations/current/Foo.table` still exists.

Electron100:
This is a good segue into a significant design change I've been wanting opinions on. One of the marquee features of Butane is the ability to have fully automatic migrations, but I unfortunately made a key architectural error in the design and implementation of this. All of the metadata to support the migration is written out by a proc-macro run during compilation. This has two major downsides:
1. Since proc macros see only the AST and not the type system, all of the `ToSql`/`FromSql` implementations are worthless here, we're essentially maintaining two separate systems of knowledge for how to map between Rust types and SQL types -- one in the type system and the other in the proc macro. This makes `butane_type` a necessary kludge when also implementing the to/from conversions and requires adding (error-prone) type name recognition to the proc macro each time we add a new core type.
2. The proc macro runs independently on each `#[model]`, and Cargo does not appear to provide any environment variables with a unique identifier for the build. Thus each `.table` file is generated independently and there isn't a great way to reap dead ones (that I've identified).

I'd like to solve the first problem and part of the second by changing when the migration metadata happens, at the cost of *slightly* more work for the user (developer using Butane). Rather than creating the migration metadata (`ADB`) at build time from the `#[model` proc macro, the proc macro would just create model implementations capable of generating migration metadata (each model could generate its corresponding `ATable`). The proc macro would also generate a program: `butane_migration_helper.rs` designed to be invoked by `butane_cli`. We'd probably need the the user to add this as a `[[bin]]` to their `Cargo.toml`. We could automate that with `butane_cli` using `toml_edit`, but would probably still want to prompt the user before making automatic changes to their project file.

We'd probably also need to introduce a new proc macro where the user has to explicitly list all of their models to fully solve the dead-model problem and make generation of `butane_migration_helpers.rs` easier otherwise we'd have no canonical point for creating it.

I don't love the two extra steps for the user, but I'd really love to fix the current lack of type system information when generating the schema, as well as the dead types issue here.

Thoughts?

jayvdb:
[...]

PR #46, opened 2023-02-06
Add a `butane clean` command

Electron100:
Clean current migration state. Deletes the current migration working state which is generated on each build. This can be used as a workaround to remove stale tables from the schema, as Butane does not currently auto-detect model removals. The next build will recreate with only tables for the extant models.

This is a workaround for #44

### DIFF
--- a/butane_cli/src/main.rs
+++ b/butane_cli/src/main.rs
@@ -17,6 +17,7 @@ fn main() {
         .version(env!("CARGO_PKG_VERSION"))
         .author("James Oakley <james@electronstudio.org>")
         .about("Manages butane database migrations")
+        .max_term_width(80)
         .subcommand(
             clap::SubCommand::with_name("init")
                 .about("Initialize the database")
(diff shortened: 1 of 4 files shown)


jayvdb:
:+1:

It would be good to call this from within the test harness, so each test module doesnt have all of the migrations of all of the tests.  Ideally it is run before the macros in the tests.  Running it afterwards (in the db shutdown hook?) would delete potentially useful data, but I guess that doesnt matter much if the tests passed.

Electron100:
> It would be good to call this from within the test harness, so each test module doesnt have all of the migrations of all of the tests

Maybe I'm being obtuse, but I don't think that will work. In my observation, within a given project `cargo test` builds all test binaries before executing any of them. Which is perhaps an argument for decoupling schema-generation-time from proc-macro-execution-time -- if they were decoupled then we *could* do as you suggest.

jayvdb:
Of course you are right. Sorry for the distraction.

jayvdb (approved):

PR #48, opened 2023-02-08
Delete tests .butane directories

jayvdb:


butane/build.rs
```diff
@@ -0,0 +1,9 @@
+fn main() {
+    let dirs = [".butane/", "../example/.butane"];
```
Electron100, line 2: Why is the build in this directory responsible for cleaning up the `.butane` directory in the example directory? It feels to me as though if cleanup is going to happen, each project should be responsible for itself.

Electron100, line 2: Could you add a comment in the build.rs giving some background for future readers on why the cleanup is happening?

jayvdb: There is a tidy perf hit by having a `build.rs`, so I was avoiding having one for each set of tests.

But I have no problem splitting it.  And adding the docs to explain why is a good idea.

Issue #20, opened 2022-12-14
Support for simple enums of a single type


jayvdb:
afaics, enums consisting of a single type are not yet automatically supported.  I can implement `ToSql`, but it would be nice to not need to.  I see there is a `#[butane_type(Text)]`, but it doesnt appear to help here.  I have serde `Serialize` and `Deserialize` for the enum, and `ToString`/`FromStr`, but I get the error below.

```rust
    #[derive(
        Clone,
        Copy,
        Debug,
        Deserialize,
        Eq,
        Hash,
        Ord,
        PartialEq,
        PartialOrd,
        Serialize,
    )]
    pub enum FormatEnum {
        #[serde(rename = "tif")]
        Tif,
        #[serde(rename = "png")]
        Png,
    }

    impl ToString for FormatEnum {
        fn to_string(&self) -> String {
            match *self {
                Self::Tif => "tif".to_string(),
                Self::Png => "png".to_string(),
            }
        }
    }

    impl std::str::FromStr for FormatEnum {
        type Err = &'static str;
        fn from_str(value: &str) -> Result<Self, Self::Err> {
            match value {
                "tif" => Ok(Self::Tif),
                "png" => Ok(Self::Png),
                _ => Err("invalid value"),
            }
        }
    }

    impl Default for FormatEnum {
        fn default() -> Self {
            FormatEnum::Tif
        }
    }
```
error:
```rust
error[E0277]: the trait bound `FormatEnum: ToSql` is not satisfied
   --> rust/models/src/lib.rs:388:5
    |
388 |     #[model]
    |     ^^^^^^^^ the trait `ToSql` is not implemented for `FormatEnum`
    |
    = help: the following other types implement trait `ToSql`:
              ...
            and 32 others
    = note: required for `std::option::Option<FormatEnum>` to implement `ToSql`
    = note: required for `SqlVal` to implement `From<std::option::Option<FormatEnum>>`
    = note: required for `std::option::Option<FormatEnum>` to implement `Into<SqlVal>`

```

enums are mentioned at https://github.com/Electron100/butane/blob/master/notes.org#roadmap

jayvdb:
fwiw, diesel also doesnt automatically support enums.
https://github.com/terry90/diesel-enum-derive and similar are required to avoid writing boilerplate code.

PR #49, opened 2023-02-08
WIP: Add #[derive(ButaneJson)]

jayvdb:
This makes it easy to mark a struct with `#[butane_type(Json)` and `#[derive(ButaneJson)]` , and then use the struct in a model.

There are likely to be much better ways of doing this.

### DIFF
--- a/butane_codegen/src/lib.rs
+++ b/butane_codegen/src/lib.rs
@@ -167,3 +167,39 @@ fn migrations_dir() -> PathBuf {
     dir.push("migrations");
     dir
 }
+
+#[proc_macro_derive(ButaneJson)]
+pub fn derive_butane_json(input: TokenStream) -> TokenStream {
+    let ast = syn::parse_macro_input!(input as syn::DeriveInput);
+    let struct_name = &ast.ident;
+    let s = format!(
+        "impl ToSql for {struct_name}
+{{
+    fn to_sql(&self) -> SqlVal {{
+        self.to_sql_ref().into()
+    }}
+    fn to_sql_ref(&self) -> SqlValRef<'_> {{
+        SqlValRef::Json(serde_json::to_value(self).unwrap())
+    }}
+}}
+
+impl FromSql for {struct_name}
+{{
+    fn from_sql_ref(val: SqlValRef) -> Result<Self, butane::Error> {{
+        if let SqlValRef::Json(v) = val {{
+            return Ok({struct_name}::deserialize(v).unwrap());
+        }}
+        Err(butane::Error::CannotConvertSqlVal(
+            SqlType::Json,
+            val.into(),
+        ))
+    }}
+}}
+impl FieldType for {struct_name}
+{{
+    type RefType = Self;
+    const SQLTYPE: SqlType = SqlType::Json;
+}}"
+    );
+    s.parse().unwrap()
+}
--- a/butane_codegen/src/lib.rs
+++ b/butane_codegen/src/lib.rs
@@ -3,7 +3,12 @@
 
 extern crate proc_macro;
 
-use butane_core::*;
+use butane_core::SqlType;
+use butane_core::{
+    codegen, make_compile_error, migrations,
+    migrations::adb::{DeferredSqlType, TypeIdentifier, TypeKey},
+    migrations::{Migration, MigrationsMut},
+};
 use proc_macro::TokenStream;
 use proc_macro2::TokenStream as TokenStream2;
 use proc_macro2::TokenTree;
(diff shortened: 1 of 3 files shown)


butane_codegen/src/lib.rs
```diff
@@ -167,3 +167,39 @@ fn migrations_dir() -> PathBuf {
     dir.push("migrations");
     dir
 }
+
+#[proc_macro_derive(ButaneJson)]
```
jayvdb, line 171: If this approach is the right one for now, the name `ButaneJson` needs to change as the `Butane..` is redundant. The new naming should allow for another derive macros, such as for enums https://github.com/Electron100/butane/issues/20

One alternative is to have something like `#[butane_type(Json, derive=true)` automatically do this, and `#[butane_type(Text, derive=true)]` would be the equivalent for enums.  This at least avoids a bunch of new names.  But is `butane_type` going to survive after https://github.com/Electron100/butane/commits/migration_v2 ?

Happy to try re-implementing this any other way that is suggested.

butane_codegen/src/lib.rs
```diff
@@ -167,3 +167,39 @@ fn migrations_dir() -> PathBuf {
...
+pub fn derive_butane_json(input: TokenStream) -> TokenStream {
+    let ast = syn::parse_macro_input!(input as syn::DeriveInput);
+    let struct_name = &ast.ident;
+    let s = format!(
```
Electron100, line 175: Any reason not to use the `quote!` macro (as the other proc macros in this crate do) rather than parsing a string? 
https://crates.io/crates/quote

butane_codegen/src/lib.rs
Electron100: I think conceptually what we're doing here is deriving `FieldType` (and its dependencies `ToSql`/`FromSql`) to enable a type to be used in model fields. However, it's conceptually a parameterized derive (Json is the parameter) which Rust doesn't really have a concept of today. So I might be inclined to define this as an attribute macro (because it's more flexible than a derive macro) something like `#[field_type(Json)]`, or maybe `#DeriveFieldType(Json)`.

It's arguably clearer to be able to write `#[Derive(..)]` here, but since `ButaneJson` isn't a real trait, I'm not sure `#[Derive(ButaneJson)]` would really help clarity

I'm not one to get super hung up on names though.

`butane_type` is not expected to survive migration v2. I've filed #60 to track that issue/experiment more clearly

jayvdb: I have updated this to be a derive of `FieldType` - it currently only handles `butane_type` of `Json` but it could now be expanded to handle other types.

Electron100:
Sorry for the slow reply here, but I'm not sure I'm wild about loading up the current migration to check whether there's been a `butane_type` association for for JSON. Since we're supporting only JSON, perhaps it would be more straightforward and foolproof to just automatically make that binding at the same time.
Like this:
https://github.com/Electron100/butane/commit/fefe26138ea6c0cf56085d50453bb2e3e94a9274

jayvdb:
I anticipated this `FieldType` derive would be extended to work for enums, and other "custom" types that are support for popular third party libs.  The restriction on only supporting `Json` is just to show the concept quickly.

Yes there is a dependency on `butane_type`, but that is mandatory at the moment.  So the question is whether your intended replacement for `butane_type` will be able to support this as well?

jayvdb:
But I am happy to go with your approach, if it means we get this merged soon.
We can worry about expanding this derive later.

Electron100:
>> I anticipated this FieldType derive would be extended to work for enums, and other "custom" types

Agreed. I think it can be extended to enums without any additional syntax. If we end up with cases where it's ambiguous what representation should be used we can introduce helper-attributes to the derive to allow further specification.

>> So the question is whether your intended replacement for butane_type will be able to support this as well?

Yes. My hope is for `butane_type` to go away entirely if I succeed in no longer having two separate views of the db type <=> Rust type mapping (the runtime one with real type info and the proc-macro one based solely on the AST the macros see). I'll admit aside from a shortage of time I've been a bit stymied by how to still preserve a good Butane CLI experience that doesn't require custom work in the codebase to help it. 
Regardless of the specifics there, `butane_type` is a crutch -- the goal is to specify the type used in the `ToSql/FromSql` impls at the same time as the type used for building the migration. So if we have an opportunity to do that with the `FieldType` derive, I think we should take it.

>> I am happy to go with your approach, if it means we get this merged soon.

Yup. I'm always open to being convinced I'm wrong, but if you're on board I'll merge your change + my amendment.

jayvdb:
> if you're on board I'll merge your change + my amendment.

Sounds good. Please do that.

Electron100:
Merged in #68

PR #54, opened 2023-02-25
More flexibility in accepting DateTime identifiers

### LINKED ISSUE #53: chrono::DateTime doesnt work with migrations
https://github.com/Electron100/butane/issues/19 added support for `chrono::DateTime`.

> Encountered unexpected error: Cannot resolve type CustomType(chrono :: DateTime < chrono :: offset :: Utc >). Are you missing a #[butane_type] attribute?

It looks like `chrono::naive::NaiveDateTime` has some implementation in `butane_core/src/db` which does not yet exist for `chrono::DateTime`, so I will see if I can add it.

We also need to work out how to ensure the tests catch this in future.

Electron100:
Allows `chrono::DateTime` alongside `DateTime`
Fixes #53

The real issue here is that the migration type resolution is relying on the proc macro, which sees only the AST and not the type system. So when it sees a type identifier, it sees the path segments that are in the AST, not the actual type, meaning that `chrono::DateTime` and `DateTime` (after `use chrono::DateTime`) are not the same! The latter was getting accepted and the former is not.

This PR is a bandaid to discard the path information and just look at the last segment (hoping it was imported from chrono!) to make this slightly more flexible.

This type of issue is the biggest reason I want to move migration generation from proc-macro time to something triggered by the `butane` cli command.

### DIFF
--- a/butane/tests/basic.rs
+++ b/butane/tests/basic.rs
@@ -98,6 +98,7 @@ struct TimeHolder {
     pub id: i32,
     pub naive: NaiveDateTime,
     pub utc: DateTime<Utc>,
+    pub utc2: chrono::DateTime<Utc>,
 }
 
 fn basic_crud(conn: Connection) {
@@ -336,6 +337,7 @@ fn basic_time(conn: Connection) {
         id: 1,
         naive: now.naive_utc(),
         utc: now,
+        utc2: now,
         state: ObjectState::default(),
     };
     time.save(&conn).unwrap();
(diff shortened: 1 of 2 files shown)


jayvdb:
I'll test this on my project which is using `pub some_date: Option<chrono::DateTime<chrono::offset::Utc>>,` 

In order to verify that these type of changes work with migration system, it would be useful to use the type in the `example/` project, which now has tests that run `makemigration`.

jayvdb (approved):
This enables `makemigration` for my scenario.

Electron100:
> In order to verify that these type of changes work with migration system, it would be useful to use the type in the example/ project,

The tests under `butane/tests` are fit for this purpose as well. The test setup in `setup_db` from `butane/tests/common/mod.rs` creates a migration, and the change this PR made to `butane/tests/basic.rs` was sufficient to trigger the failure prior to the fix. The tests did *previously* include a `DateTime`, with a previous `use chrono::DateTime` but not the qualified-in-place `chrono::DateTime`. While these are semantically the same, they aren't the same in the AST the proc macro sees.

PR #55, opened 2023-03-01
Add Debug trait to all data structures

jayvdb:
Also alpha sorts derives.


butane_core/src/db/sqlite.rs
```diff
@@ -363,6 +366,7 @@ impl<'a> QueryAdapterInner<'a> {
     }
 }
 
+// Debug can not be derived because QueryAdapterInner above doesn't implement it.
```
jayvdb, line 369: my guess is this struct and `QueryAdapterInner` above wont be encountered very often by developers, of butane or butane apps, so wasnt worth adding custom impls.  But I dont mind doing them for completeness.

Electron100 (approved):

PR #57, opened 2023-03-01
Support fake data

jayvdb:


butane_core/src/fkey.rs
```diff
@@ -199,3 +202,11 @@ where
         Ok(Self::from_pk(T::PKType::deserialize(deserializer)?))
     }
 }
+
+#[cfg(feature = "fake")]
+/// Fake data support is currently limited to empty ForeignKey relationships.
```
Electron100, line 207: When would this actually be useful? Since this creates an invalid `ForeignKey` (not fully initialized) and most of the methods will panic if called

Electron100: I guess the idea is that none of the methods would get called since you wouldn't use it with a real db connection given that it's fake data?

jayvdb: Ya, at the moment the `ForeignKey`, and `Many`, are not well supported, and the user needs to know what they are doing to manually help the fake-ing process like I did in the fake test cases.  Goal is for that to be automatic eventually, but at the moment this PR allows butane and fake to co-exist on the same data structures.

Issue #58, opened 2023-03-02
Can't use Many in renamed table


joaommartins:
When renaming a table like with the `$[table = ""]` attribute, it is not possible to follow many-to-many relationships from this table, as the load procedure ignores the renaming of the table and attempts to load the automatically many-to-many table using the struct name.

Migrations generation is not affected, since it correctly carries the renaming, generating the correct table for the renamed struct, as well for the many-to-many table.

joaommartins:
Fixed by #59

PR #59, opened 2023-03-02
Fixes issue with many to many relationship not respecting renamed tables

joaommartins:
# What
Fix for #58.

Adds test to cover issue, as well as a minor update to the documentation regarding the generated many-to-many table name.


Electron100 (approved):
Thank you for fixing this!

PR #61, opened 2023-03-21
Increase test coverage for butane_code::db

jayvdb:
- `butane_code::db::*` increases for 76.75 % to 81.49 %, but the big wins are mod.rs goes up 7% (now ~95%), and `pg.rs` goes up ~10% (now 76.3%, from red to yellow in `grcov`'s html output).
- The separation of the test helper also allows using it for test cases in project's using butane.
- And it seems running the tests is a bit faster.


jayvdb:
fwiw, we are using https://deepsource.io/ for Rust projects, including code coverage tracking, and it is pretty good.  If you enable this project in deepsource, or codecov, I can do the GitHub Actions work to send the coverage data.

Electron100 (approved):

Electron100:
Thanks. I'm merging this and will try to look into deepsource.io.

Issue #65, opened 2023-04-20
Move butane_cli non-cli specific code to a lib


jayvdb:
I would like to programmatically run the operations in `butane_cli` .  e.g. `make_migration` needs to be part of our build process, and ideally be run from within `build.rs`.  This can be done using only `butane`, however there are a bunch of utilities and `struct CliState` which need to be duplicated to achieve this.

The simplest approach is to move these to a `lib.rs` within `butane_cli` , and document that anything `pub` in it is very volatile.  That way these dont become part of the more stable `butane` API.  We can also use some rustdoc specific attributes to hide them from the crate.io docs, to make it even less likely people will use them without being aware of the risks involved.

This will also make it easier to write tests for the cli.

PR #66, opened 2023-04-20
Create butane_cli lib.rs

jayvdb:
Closes https://github.com/Electron100/butane/issues/65

### DIFF
--- a/butane_cli/src/lib.rs
+++ b/butane_cli/src/lib.rs
@@ -1,0 +1,273 @@
+#![doc(hidden)]
+//! This library is not stable, and usage is strongly discouraged.
+//!
+//! It is intended only to assist developing the CLI.
+//! Usage of this library is strongly discouraged unless you expect & accept
+//! breakages in the future.
+//! Backwards compatibility of the library will not even be considered, as the
+//! only objective of the crate is to provide a stable CLI.
+use std::{fs::File, io::Write, path::PathBuf};
+
+use butane::migrations::{
+    copy_migration, FsMigrations, MemMigrations, Migration, MigrationMut, Migrations, MigrationsMut,
+};
+use butane::query::BoolExpr;
+use butane::{db, db::Connection, db::ConnectionMethods, migrations};
+use chrono::Utc;
+use serde::{Deserialize, Serialize};
+
+pub type Result<T> = std::result::Result<T, anyhow::Error>;
+
+#[derive(Serialize, Deserialize, Default)]
+pub struct CliState {
+    embedded: bool,
+}
+impl CliState {
+    pub fn load() -> Result<Self> {
+        let path = base_dir()?.join("clistate.json");
+        let file = File::open(path);
+        match file {
+            Ok(file) => Ok(serde_json::from_reader(file)?),
+            Err(_) => Ok(CliState::default()),
+        }
+    }
+
+    pub fn save(&self) -> Result<()> {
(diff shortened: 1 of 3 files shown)


butane_cli/src/lib.rs
```diff
@@ -0,0 +1,265 @@
...
+}
+
+pub fn default_name() -> String {
+    Utc::now().format("%Y%m%d_%H%M%S%3f").to_string()
+}
+
+pub fn init(name: &str, connstr: &str) -> Result<()> {
+    if db::get_backend(name).is_none() {
+        eprintln!("Unknown backend {name}");
+        std::process::exit(1);
```
Electron100, line 42: How are you thinking about error handling for the library usage? Exiting the process is probably not desirable

butane_cli/src/lib.rs
```diff
@@ -0,0 +1,265 @@
...
+    }
+}
+
+pub fn default_name() -> String {
+    Utc::now().format("%Y%m%d_%H%M%S%3f").to_string()
+}
+
+pub fn init(name: &str, connstr: &str) -> Result<()> {
+    if db::get_backend(name).is_none() {
+        eprintln!("Unknown backend {name}");
```
Electron100, line 41: Should we switch to some sort of logging facade (log or trace) here? I'm not sure direct printing makes sense for library usage

butane_cli/src/lib.rs
```diff
@@ -0,0 +1,265 @@
...
+        Ok(spec) => Ok(spec),
+        Err(butane::Error::IO(_)) => {
+            eprintln!("No Butane connection info found. Did you run butane init?");
+            std::process::exit(1);
+        }
+        Err(e) => Err(e.into()),
+    }
+}
+
+pub fn list_migrations() -> Result<()> {
```
Electron100, line 172: This just prints to the console rather than returning anything, so I'm not sure how it's useful for programmatic usage? Perhaps I'm misunderstanding how you're envisioning using this.

Electron100 (commented):
I'm on board with the general idea. I'd be interested in hearing more about your programmatic usage though. There are a few things moved to the lib.rs that don't seem suitable for general programmatic usage (printing to the console, exiting the process on errors)

jayvdb:
The console prints are desirable to me.  The `exit` probably isnt ideal, but it is reasonable for my use case.

My usage is driving multiple "cli" steps from a utility to ensure all necessary steps and checks are done, avoiding reliance on developers doing the process and allowing CI to fail if the steps havent been done.  for me, failure during the cli steps is appropriate to cause my utility to also fail.

But this is only a very quick WIP so far, and can be improved to be more useful to a wider audience.  However I dont think it needs to be over-engineered, as I think it is more important that this PR does the work needed to strongly discourage anyone from using the lib unless they expect & accept breakages in the future due to the highly volatile nature of the inner workings of this crate, since its primary objective is to be a CLI , not a lib.

Electron100:
Sounds good. The boundary feels a bit hacky/unprincipled but as long as it's being commented unstable I'm on board.

jayvdb:
@Electron100 , are you happy for this to go in as-is, or do you want me to do some of the lib-ification that you mentioned above, all of which I agree with, but I think they can be done only when the cli needs cause those types of improvements to be necessary

jayvdb:
Now it is published, we can see it worked at https://docs.rs/butane_cli/latest/butane_cli/

PR #72, opened 2023-05-02
Derive FieldType for simple enums

### LINKED ISSUE #20: Support for simple enums of a single type

Electron100:
Support automatic deriving of `FieldType` for simple enums. The database representation is the string form of the variant name. This builds on the capability from #68 

Fixes #20

Future work could allow an option to use discriminant value instead of name.

### DIFF
--- a/butane_codegen/src/lib.rs
+++ b/butane_codegen/src/lib.rs
@@ -175,16 +175,20 @@ pub fn derive_field_type(input: TokenStream) -> TokenStream {
     match derive_input.data {
         syn::Data::Struct(_) => derive_field_type_with_json(ident),
         syn::Data::Enum(data_enum) => derive_field_type_for_enum(ident, data_enum),
-        syn::Data::Union(_) => panic!("Cannot derive FieldType for a union"),
+        syn::Data::Union(_) => derive_field_type_with_json(ident),
     }
 }
 
 fn derive_field_type_for_enum(ident: &Ident, data_enum: syn::DataEnum) -> TokenStream {
-    data_enum.variants.iter().for_each(|variant| {
-        if variant.fields != syn::Fields::Unit {
-            panic!("Only simple enums are supported when deriving FieldType")
-        }
-    });
+    if data_enum
+        .variants
+        .iter()
+        .any(|variant| variant.fields != syn::Fields::Unit)
+    {
+        // Non-simple enum, fall back to json derive
+        return derive_field_type_with_json(ident);
+    }
+
     let match_arms_to_string: Vec<TokenStream2> = data_enum
         .variants
         .iter()
--- a/butane_codegen/src/lib.rs
+++ b/butane_codegen/src/lib.rs
@@ -170,8 +170,82 @@ fn migrations_dir() -> PathBuf {
 
 #[proc_macro_derive(FieldType)]
 pub fn derive_field_type(input: TokenStream) -> TokenStream {
-    let ast = syn::parse_macro_input!(input as syn::DeriveInput);
-    derive_field_type_with_json(&ast.ident)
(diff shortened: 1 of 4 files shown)


Electron100:
@jayvdb does this meet your needs for #20?

jayvdb (approved):

butane_codegen/src/lib.rs
```diff
@@ -170,8 +170,82 @@ fn migrations_dir() -> PathBuf {
...
 #[proc_macro_derive(FieldType)]
 pub fn derive_field_type(input: TokenStream) -> TokenStream {
-    let ast = syn::parse_macro_input!(input as syn::DeriveInput);
-    derive_field_type_with_json(&ast.ident)
+    let derive_input = syn::parse_macro_input!(input as syn::DeriveInput);
+    let ident = &derive_input.ident;
+    match derive_input.data {
+        syn::Data::Struct(_) => derive_field_type_with_json(ident),
+        syn::Data::Enum(data_enum) => derive_field_type_for_enum(ident, data_enum),
+        syn::Data::Union(_) => panic!("Cannot derive FieldType for a union"),
```
jayvdb, line 178: can we have this failure mode instead revert back to using `derive_field_type_with_json` ?

butane_codegen/src/lib.rs
```diff
@@ -170,8 +170,82 @@ fn migrations_dir() -> PathBuf {
...
+    }
+}
+
+fn derive_field_type_for_enum(ident: &Ident, data_enum: syn::DataEnum) -> TokenStream {
+    data_enum.variants.iter().for_each(|variant| {
+        if variant.fields != syn::Fields::Unit {
+            panic!("Only simple enums are supported when deriving FieldType")
```
jayvdb, line 185: can we have this failure mode instead revert back to using `derive_field_type_with_json` (if that feature is enabled) which does support complex enums?

PR #77, opened 2023-05-05
Bump fake and rusqlite deps

jayvdb:


jayvdb:
This bumping of fake is very important for us, as the new version has https://github.com/cksac/fake-rs/pull/130 and https://github.com/cksac/fake-rs/pull/129 included, both of which we use, and we'd like to discontinue publishing our own fork of that package now.

PR #84, opened 2023-05-10
Fix publishing

Electron100:
Fix publishing with `cargo release`
1. Fix version specifications for in-repo packages in workspace Cargo.toml
2. Don't release butane_test_helper
3. Don't specify version for `butane_test_helper` when referencing it. This allows `butane_core` to be published despite the circular dev-dependency. See https://github.com/rust-lang/cargo/pull/7333


jayvdb (approved):
lgtm. I am publishing `butane_test_helper` and using it in our project as a test helper, but are already adjusting its `Cargo.toml` before publishing.

PR #90, opened 2023-05-19
Update notes.org

jayvdb:


notes.org
```diff
@@ -37,24 +37,29 @@
...
 ** vNext
-	 + foreign key constraints in db, cascade setting
+	 + foreign key constraints in db, cascade setting [[https://github.com/Electron100/butane/issues/21]]
 	 + incremental save
 	 + Many backreferences
-	 + Column renames in migrations
+	 + Column renames in migrations [[https://github.com/Electron100/butane/issues/89]]
 	 + prepared/reusable queries
 	 + Improved custom support, including raw sql where clauses
 	 + More precise spans in macro errors for invalid field names
```
jayvdb, line 60: These are likely a few easy improvements.  Do we know some instances that need improvement?

notes.org
```diff
@@ -37,24 +37,29 @@
 		 let posts_in_blog = query!(Post, &conn, blog = {blog}).load(&conn)
 	 }
 	 #+END_SRC
+
 	 struct Posts
+
+	 #+BEGIN_SRC sql
```
jayvdb, line 43: do we need this, or the Rust above?

i.e. is anything in the above missing from butane?

notes.org
```diff
@@ -37,24 +37,29 @@
...
-	 + foreign key constraints in db, cascade setting
+	 + foreign key constraints in db, cascade setting [[https://github.com/Electron100/butane/issues/21]]
 	 + incremental save
 	 + Many backreferences
-	 + Column renames in migrations
+	 + Column renames in migrations [[https://github.com/Electron100/butane/issues/89]]
 	 + prepared/reusable queries
 	 + Improved custom support, including raw sql where clauses
 	 + More precise spans in macro errors for invalid field names
 	 + Remove timestamp from migration names
```
jayvdb, line 61: I quite like the timestamp being in the name, but it is problematic when `git` rebasing a change.  I can write up an issue about this, but I probably wont be able to fully capture the original intent of this item.

notes.org
```diff
@@ -37,24 +37,29 @@
...
 * Roadmap
 ** vNext
-	 + foreign key constraints in db, cascade setting
+	 + foreign key constraints in db, cascade setting [[https://github.com/Electron100/butane/issues/21]]
 	 + incremental save
 	 + Many backreferences
-	 + Column renames in migrations
+	 + Column renames in migrations [[https://github.com/Electron100/butane/issues/89]]
 	 + prepared/reusable queries
 	 + Improved custom support, including raw sql where clauses
```
jayvdb, line 59: "raw sql where clauses" is a reasonable clear enhancement.  Im not sure what else falls under "Improved custom support"

notes.org
```diff
@@ -37,24 +37,29 @@
...
+
 * Roadmap
 ** vNext
-	 + foreign key constraints in db, cascade setting
+	 + foreign key constraints in db, cascade setting [[https://github.com/Electron100/butane/issues/21]]
 	 + incremental save
 	 + Many backreferences
-	 + Column renames in migrations
+	 + Column renames in migrations [[https://github.com/Electron100/butane/issues/89]]
 	 + prepared/reusable queries
```
jayvdb, line 58: IMO this isnt a limitation - there are some performance gains here, but ... I doubt typical users of `butane` would see this as a priority.

notes.org
```diff
@@ -37,24 +37,29 @@
...
+		body TEXT NOT NULL,
+		published BOOLEAN NOT NULL DEFAULT 'f'
+	 )
+	 #+END_SRC
+
 * Roadmap
 ** vNext
-	 + foreign key constraints in db, cascade setting
+	 + foreign key constraints in db, cascade setting [[https://github.com/Electron100/butane/issues/21]]
 	 + incremental save
```
jayvdb, line 55: This one isnt clear to me.
does this mean doing writes immediately when changing a struct instance?  sounds crazy.
or `Many.add` writing immediately?

notes.org
```diff
@@ -37,24 +37,29 @@
...
 	 + Many backreferences
```
jayvdb, line 56: I guess this is simply a function to take the caller from a `Many` instance to the instance that "(owner table, key)" points to?
Seems so, and seems not too hard to do, and also quite useful.

notes.org
jayvdb: https://github.com/Electron100/butane/issues/91

notes.org
Electron100: No, this is obsolete and can be deleted

notes.org
Electron100: This is probably best covered by issue #1  now

notes.org
Electron100: The original intention of this line was `save` being able to do an update with only the fields that have actually changed, as a performance enhancement. However that really doesn't fit well with the current architecture of assigning directly to struct fields and I'm not sure I think it's a good idea any more. I think at this point I'd do something like `DataResult` (or like what Diesel does) to allow updates with a struct containing only a subset of the model fields.

notes.org
Electron100: Correct

Issue #63, opened 2023-04-19
Find .butane in workspace


jayvdb:
If `butane` cli is run at the top level of a cargo workspace, and the `.butane` is under a project within the workspace, the cli fails with

> No Butane connection info found. Did you run butane init?

It would be nice if it detected the workspace, and listed the projects with an existing `.butane`, and told the user to select one of them, with a CLI arg to specify the path to use as the base.

PR #98, opened 2023-05-29
Find .butane in workspace

jayvdb:
Closes https://github.com/Electron100/butane/issues/63

### DIFF
--- a/butane_cli/src/lib.rs
+++ b/butane_cli/src/lib.rs
@@ -17,6 +17,7 @@ use butane::migrations::{
 };
 use butane::query::BoolExpr;
 use butane::{db, db::Connection, db::ConnectionMethods, migrations};
+use cargo_metadata::MetadataCommand;
 use chrono::Utc;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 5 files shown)


butane_cli/src/lib.rs
```diff
@@ -259,10 +261,46 @@ pub fn get_migrations() -> Result<FsMigrations> {
     Ok(migrations::from_root(root))
 }
 
+/// The members of a workspace are given a PackageId, which contains the directory
+/// name however it needs to be extracted.
+pub fn extract_package_directory(
```
Electron100, line 266: I'm not a big fan of of using `package_id.repr` because it's documented to be unstable. I tried out the following locally and it seems to work. The key point here is that we can use the `PackageId` to match against the list of `Package`. There's a lot more info on `Package` including the path to the project's `Cargo.toml`
```
pub fn extract_package_directory(
    packages: &Vec<cargo_metadata::Package>,
    package_id: cargo_metadata::PackageId,
) -> Result<std::path::PathBuf> {
    let pkg = packages
        .iter()
        .find(|p| p.id == package_id)
        .ok_or(anyhow!("No package found"))?;
    let parent = pkg.manifest_path.parent().unwrap();
    Ok(parent.to_owned().into())
}
```

butane_cli/src/lib.rs
```diff
@@ -259,10 +261,46 @@ pub fn get_migrations() -> Result<FsMigrations> {
...
+        let metadata = MetadataCommand::new().no_deps().exec()?;
+        let workspace_members = metadata.workspace_members;
+
+        // Find the first workspace member with a .butane
+        for member in workspace_members {
+            let package_dir = extract_package_directory(member)?;
+            let member_butane_dir = package_dir.join(".butane/");
+
+            if member_butane_dir.exists() {
+                eprintln!("Using workspace member {:?}", member_butane_dir);
```
Electron100, line 296: I think this should go to stdout rather than stderr as it's informational, not fatal/important.

butane_cli/src/lib.rs
```diff
@@ -259,10 +261,46 @@ pub fn get_migrations() -> Result<FsMigrations> {
...
+/// Find a .butane directory to act as the base for butane.
 pub fn base_dir() -> Result<PathBuf> {
-    std::env::current_dir()
-        .map(|d| d.join(".butane"))
-        .map_err(|e| e.into())
+    let current_directory = std::env::current_dir()?;
+    let local_butane_dir = current_directory.join(".butane/");
+
+    if !local_butane_dir.exists() {
+        let metadata = MetadataCommand::new().no_deps().exec()?;
```
Electron100, line 287: We should handle the error here rather than just propagating it. If I try to run `butane` in a directory with no Cargo project/workspace today, it prints "No Butane connection info found. Did you run butane init?". That message could arguably be improved, but it at least takes a stab at explaining the problem and offering a mitigation.

With this change it prints only "Encountered unexpected error: `cargo metadata` exited with an error: error: could not find `Cargo.toml` in `/home/james/projects/mood` or any parent directory"

butane_cli/src/lib.rs
```diff
@@ -259,10 +261,46 @@ pub fn get_migrations() -> Result<FsMigrations> {
...
+        let workspace_members = metadata.workspace_members;
+
+        // Find the first workspace member with a .butane
```
Electron100, line 290: I'm a bit uneasy about just picking the first one without any further interaction. For one thing I don't think it's even stable. I tried running it from a workspace containing a couple different projects both with butane set up and while I mostly got the same project, sometimes I got a different one.

I agree it's a nice quality-of-life improvement to be able to run `butane` from the workspace root, and if there's only one workspace project with a `.butane` directory it makes sense to pick that one. But I think we need to check whether there's more than one, and not proceed without confirmation if there are choices. We can offer a flag (like cargo's `-p`) to select the project to operate on.

If we're in a subdirectory of a project it probably makes sense to automatically select that project since the working directory is giving us a hint about what the user intended.

Electron100 (changes requested):

butane_cli/src/lib.rs
```diff
@@ -263,10 +264,63 @@ pub fn get_migrations(base_dir: &Path) -> Result<FsMigrations> {
...
+/// Find a .butane directory to act as the base for butane.
 pub fn base_dir() -> Result<PathBuf> {
-    std::env::current_dir()
-        .map(|d| d.join(".butane"))
-        .map_err(|e| e.into())
+    let current_directory = std::env::current_dir()?;
+    let local_butane_dir = current_directory.join(".butane/");
+
+    if !local_butane_dir.exists() {
+        if let Ok(member_butane_dir) = find_butane_dir_in_workspace() {
```
jayvdb, line 316: errors in related to running outside of a cargo project are now ignored here.  They are silently ignored, and that could be improved, but I would like to add `-p`/`--path` in a separate PR, and then the error messages can be improved to tell the user to use those flags.

butane_cli/src/lib.rs
```diff
@@ -263,10 +264,63 @@ pub fn get_migrations(base_dir: &Path) -> Result<FsMigrations> {
...
+    Ok(parent.to_owned().into())
+}
+
+/// Find all of the `.butane` directories in the cargo workspace
+pub fn find_butane_dirs_in_workspace() -> Result<Vec<PathBuf>> {
+    let metadata = MetadataCommand::new().no_deps().exec()?;
+    let workspace_members = metadata.workspace_members;
+
+    let mut possible_directories: Vec<PathBuf> = vec![];
+    // Find the first workspace member with a .butane
```
Electron100, line 287: This comment is outdated, this method now returns all, not the first

Electron100 (approved):

Issue #91, opened 2023-05-19
Rebasing migrations


jayvdb:
One of the items in https://github.com/Electron100/butane/blob/877b65f/notes.org is "Remove timestamp from migration names".  This refers to the directory name for the migration which appears in `.butane/migrations/`, and are named like `YYMMDD_<seconds>_<migration_name>`.

I don't know all of what might be intended by that org entry, but one reason for removing timestamp from the migration name is that it problematic when it comes to changing the order of pending migrations.

i.e. two PRs are submitted that both include a migration.  The date is in the migration names.  For the purposes of simplifying the explanation, assume PR "A" is created on Monday, PR "B" is created *and* merged on Tuesday to `main` without modification, and the PR "A" is updated and merged to `main` on Wednesday.

The first obvious problem is the date of the PR "A" will be before PR "B", despite the real order being the other way around.  The date a migration was created is not helpful from the perspective of a migration.  The order that matters is which order the migration needs to be applied in.

The more difficult problem is that updating PR "A" is hard.  When PR "B" is merged, it will cause conflicts in PR "A".  This is good at present, because it prevents accidentally merging PR "A" without first rebasing it on top of PR "B".

And all of this is a very basic situation - it gets much harder if there are multiple branches, such as supporting multiple versions, and various patches being back-ported to older versions in all unpredictable order.

The approach I am using is to

1. delete the migration in PR "A" and then
2. `git checkout main .butane/migrations`, then 
3. do the git rebase, and 
4. rebuild project which will recreate the migration state on top of PR "B".
5. `butane makemigration <name>` to recreate the migration

This process means the date of the PR "A" migration will be after the date of PR "B", which makes the date prefix sensible again.

Can we improve this?  At the very least we can create a `butane_cli` command that "unmake" the last migration (delete it and rollback `.butane/migrations/state.json` to prior migration name), which is basically step (1).  The `clean` command deletes all of `.butane/migrations/current/` , but not the existing migrations, which is step (2).

The prior migration name is available in the `from_name` in the `info.json` in each migration.

I've spent a lot of time with Django, and it is seen as quite good.  We should seriously consider their migration approach unless it can be improved upon.  https://docs.djangoproject.com/en/4.2/topics/migrations/

Django default a serial number as the prefix for migration names, but does - by default - include the date in the middle of the migration name.  This reduces the importance of the date.  The serial number is just as bad as a date, as it is common to have migrations merged out of order.

[...]

PR #100, opened 2023-05-30
Detach latest migration

jayvdb:
Step one of https://github.com/Electron100/butane/issues/91 - detach rather than unmake.

### DIFF
--- a/butane_cli/src/main.rs
+++ b/butane_cli/src/main.rs
@@ -1,10 +1,10 @@
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 
 use clap::{value_parser, Arg, ArgMatches};
 
 use butane_cli::{
-    base_dir, clean, clear_data, collapse_migrations, delete_table, embed, handle_error,
-    list_migrations, migrate, Result,
+    base_dir, clean, clear_data, collapse_migrations, delete_table, detach_latest_migration, embed,
+    handle_error, list_migrations, migrate, Result,
 };
 
 fn main() {
@@ -52,12 +52,6 @@ fn main() {
         .subcommand(
             clap::Command::new("detachmigration")
                 .about("Detach the top migration")
-                .arg(
-                    Arg::new("NAME")
-                        .required(true)
-                        .index(1)
-                        .help("Name to use for the migration"),
-                ),
         )
         .subcommand(clap::Command::new("migrate").about("Apply migrations"))
         .subcommand(clap::Command::new("list").about("List migrations"))
@@ -113,9 +107,7 @@ fn main() {
         Some(("makemigration", sub_args)) => {
             handle_error(make_migration(&base_dir, Some(sub_args)))
         }
-        Some(("detachmigration", sub_args)) => {
-            handle_error(detach_migration(&base_dir, Some(sub_args)))
-        }
+        Some(("detachmigration", _)) => handle_error(detach_latest_migration(&base_dir)),
         Some(("migrate", _)) => handle_error(migrate(&base_dir)),
         Some(("rollback", sub_args)) => handle_error(rollback(&base_dir, Some(sub_args))),
         Some(("embed", _)) => handle_error(embed(&base_dir)),
@@ -149,11 +141,6 @@ fn init(base_dir: &PathBuf, args: Option<&ArgMatches>) -> Result<()> {
 fn make_migration(base_dir: &PathBuf, args: Option<&ArgMatches>) -> Result<()> {
     let name_arg = args.and_then(|a| a.get_one::<String>("NAME"));
     butane_cli::make_migration(base_dir, name_arg)
-}
-
-fn detach_migration(base_dir: &Path, args: Option<&ArgMatches>) -> Result<()> {
-    let name_arg = args.and_then(|a| a.get_one::<String>("NAME"));
-    butane_cli::detach_migration(base_dir, name_arg)
 }
 
 fn rollback(base_dir: &PathBuf, args: Option<&ArgMatches>) -> Result<()> {
--- a/butane_cli/src/main.rs
+++ b/butane_cli/src/main.rs
@@ -49,6 +49,16 @@ fn main() {
                         .help("Name to use for the migration"),
                 ),
         )
+        .subcommand(
+            clap::Command::new("detachmigration")
+                .about("Detach the top migration")
+                .arg(
+                    Arg::new("NAME")
+                        .required(true)
+                        .index(1)
+                        .help("Name to use for the migration"),
+                ),
+        )
         .subcommand(clap::Command::new("migrate").about("Apply migrations"))
         .subcommand(clap::Command::new("list").about("List migrations"))
         .subcommand(clap::Command::new("collapse").about("Replace all migrations with a single migration representing the current model state.").arg(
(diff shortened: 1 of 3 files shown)


butane_core/src/migrations/fsmigrations.rs
```diff
@@ -271,6 +271,17 @@ impl FsMigrations {
         f.write_all(serde_json::to_string(state)?.as_bytes())
             .map_err(|e| e.into())
     }
+    pub fn detach_latest_migration(&mut self) -> Result<()> {
+        let latest = self.latest().expect("There are no migrations");
```
Electron100, line 275: This line should return an error result if there are no migrations, not panic.

Electron100: Same for line 279

jayvdb: This was hastily put together to help think through how detach/re-attach might work.  The PR will be abandoned if you disagree with the suggestion at https://github.com/Electron100/butane/issues/91#issuecomment-1567598015 of allowing people to edit the `.sql` files.  Without out that, the `unmake` approach is the better approach.

Electron100: Sorry, missed this reply. I agree with allowing people to edit the .sql files -- that's always been supported. The idea is that the migration system is as good as possible, but manual editing of the migration sql should be an escape hatch.

jayvdb:
I was a bit ambiguous initially planning to allow any migration to be detached.  While detaching from the middle is do-able, it is less sensible to allow it, and it is not necessary for most common #91 scenarios.  Doing multiple detaches of the top migration is sufficient, and adding support for re-attaching migrations is more important IMO.

Electron100:
Ack on the review request. Will take a look tomorrow.

jayvdb:
Note two parts of https://github.com/Electron100/butane/issues/91#issuecomment-1567598015 that I havent implemented yet is 

> Undo it from the embed

and

> When there is an existing detached migration, all commands would be disabled except for the "rejoin", with an error message informing the user how to delete the detached migration if they want to abandon it.

The first of those is mandatory.  

IMO the second doesnt depend on "rejoin" being built first - all commands can be blocked until the detached migration has been removed - atm that means it is the users responsibility to do something with it or delete it.  I'm happy to do that in the scope of this PR if you agree that is good enough for now.

butane_cli/src/main.rs
```diff
@@ -49,6 +49,10 @@ fn main() {
                         .help("Name to use for the migration"),
                 ),
         )
+        .subcommand(
+            clap::Command::new("detachmigration")
+                .about("Detach the latest migration")
```
Electron100, line 54: Could we give a little more information here (or use `after_help`)? I'm not sure the concept of detaching a migration will be intuitive, so I think we should explain what it means to do so and why one might want to do so (a summary of what's described in #91)

butane_core/src/migrations/fsmigrations.rs
```diff
@@ -271,6 +271,24 @@ impl FsMigrations {
         f.write_all(serde_json::to_string(state)?.as_bytes())
             .map_err(|e| e.into())
     }
+    /// Detach the latest migration from the list of migrations,
+    /// leaving the migration on the filesystem.
+    pub fn detach_latest_migration(&mut self) -> Result<()> {
```
Electron100, line 276: Making this method solely an impl on `FsMigrations` rather than a trait method of `MigrationsMut` is inconsistent with how all the other operations methods from the CLI are handled. I think I'd prefer to stick with the existing pattern and make it a trait method for consistency/cleanliness. Since there are no other impls of the trait beyond the in-memory backend used by some tests though, I don't feel super strongly.

jayvdb: It was put here as `detach_latest_migration` needs to use `get_state` which only exists on `FsMigrations` and doesnt make sense in the context of `MemMigrations`

butane_cli/src/main.rs
```diff
@@ -49,6 +49,10 @@ fn main() {
                         .help("Name to use for the migration"),
                 ),
         )
+        .subcommand(
+            clap::Command::new("detachmigration")
```
jayvdb, line 53: IMO we should be using names like `detach-migration`, but I have followed the existing style.

butane_core/src/migrations/fsmigrations.rs
jayvdb: It might make more sense to rename `detach_latest_migration` to be `rollback_state`, or even `downgrade` if we want similar terminology between removing a migration from the database and removing a migration from the series on the filesystem.

jayvdb:
I've added the ability to detect if there are detached migrations, but after looking through the existing list of commands I am reconsidering whether this is needed at all.  While it might be usually a bad idea to run some of the commands when there is a detached migration, it cant cause new undesirable states which we *must* avoid, and maybe we should just let the user do the commands and let them deal with the repercussions.

What might be useful is to always print the list of detached migrations, so they are not accidentally forgotten.

butane_core/src/migrations/fsmigrations.rs
Electron100: Ah that's fair. We can revisit in the future if we end up with other "real" migration backends for some reason.

butane_cli/src/main.rs
Electron100: That's fair, it probably would be better naming. Perhaps use Clap's alias to support both? https://docs.rs/clap/latest/clap/struct.Command.html#method.alias

butane_cli/src/main.rs
```diff
@@ -113,6 +113,20 @@ However if the migration has been manually edited, it will need to be manually r
     let args = app.get_matches();
     let mut base_dir = args.get_one::<PathBuf>("path").unwrap().clone();
     base_dir.push(".butane");
+
+    // Prevent 'makemigration' when there is a detached migration.
+    match args.subcommand() {
+        Some(("makemigration", _)) => {
+            if let Ok(ms) = get_migrations(&base_dir) {
+                if ms.has_detached_migration().unwrap_or(false) {
+                    eprintln!("Found detached migration. Please delete or manually re-attach it.");
```
Electron100, line 122: Since `butane` doesn't provide any help in that, we should probably print out the file path to the detached migration so the user at least knows where it is.

butane_core/src/migrations/fsmigrations.rs
```diff
@@ -289,6 +289,21 @@ impl FsMigrations {
         self.save_state(&state)?;
         Ok(())
     }
+    /// Determines if a migration has been detached.
+    pub fn has_detached_migration(&self) -> Result<bool> {
+        let migration_series = self.all_migrations()?;
+        for entry in std::fs::read_dir(self.root.clone())? {
+            let path = entry?.path();
+            let file_name = path.file_name().unwrap();
+            if file_name == "current" || file_name == "state.json" {
```
Electron100, line 298: Rather than checking for `state.json` consider checking `path.is_dir()`. Only directories indicate migrations. Files can't be migrations themselves.

Electron100:
I agree with your reconsideration -- it doesn't seem necessary to me to actually block operations for detached migrations.

Electron100 (approved):

Issue #102, opened 2023-05-30
Column name quoting


jayvdb:
I have a field called `when`, and I am getting what looks like a column name quoting bug

```
thread '...' panicked at 'called `Result::unwrap()` on an `Err` value: SQLite(SqliteFailure(Error { code: Unknown, extended_code: 1 }, Some("near \"when\": syntax error")))', rust/models/tests/common/mod.rs:20:34
```

jayvdb:
Easy to demonstrate with
```diff
diff --git a/butane/tests/basic.rs b/butane/tests/basic.rs
index 816c58d..e69abe6 100644
--- a/butane/tests/basic.rs
+++ b/butane/tests/basic.rs
@@ -102,7 +102,7 @@ struct TimeHolder {
     pub id: i32,
     pub naive: NaiveDateTime,
     pub utc: DateTime<Utc>,
-    pub utc2: chrono::DateTime<Utc>,
+    pub when: chrono::DateTime<Utc>,
 }
 
 fn basic_crud(conn: Connection) {
@@ -343,7 +343,7 @@ fn basic_time(conn: Connection) {
         id: 1,
         naive: now.naive_utc(),
         utc: now,
-        utc2: now,
+        when: now,
         state: ObjectState::default(),
     };
     time.save(&conn).unwrap();
```

Looking into the fix now.

PR #104, opened 2023-05-30
Quote reserved SQL words

jayvdb:
Closes https://github.com/Electron100/butane/issues/102

https://github.com/Electron100/butane/pull/105 should be merged first.

Notes:
- I dont mind if you want the `String` version - commit 1. Or squash in the `Cow<str>`.
- It would be useful to add specific tests for the new function, but it is getting a *lot* of coverage.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -201,6 +201,7 @@ dependencies = [
  "rusqlite",
  "serde",
  "serde_json",
+ "sqlparser",
  "syn 2.0.15",
  "tempdir",
  "thiserror",
@@ -1544,6 +1545,15 @@ checksum = "02e2d2db9033d13a1567121ddd7a095ee144db4e1ca1b1bda3419bc0da294ebd"
 dependencies = [
  "libc",
  "winapi",
+]
+
+[[package]]
+name = "sqlparser"
+version = "0.34.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "37d3706eefb17039056234df6b566b0014f303f867f2656108334a55b8096f59"
+dependencies = [
+ "log",
 ]
 
 [[package]]
--- a/butane/tests/basic.rs
+++ b/butane/tests/basic.rs
@@ -102,7 +102,7 @@ struct TimeHolder {
     pub id: i32,
     pub naive: NaiveDateTime,
     pub utc: DateTime<Utc>,
-    pub utc2: chrono::DateTime<Utc>,
+    pub when: chrono::DateTime<Utc>,
 }
 
 fn basic_crud(conn: Connection) {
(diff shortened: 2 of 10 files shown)


Electron100:
Thank you for this fix. But why not just always quote identifiers? Then we could avoid the new dependency or having multiple cases to worry about. I haven't measured, but given the array lookup and the upcase, I would guess that trying to figure out whether a given identifier needs to be quoted or not will be slower than just always quoting, and we could put the quotes directly in the `write!` macro invocations to avoid any more allocation than is already happening.

I feel as though I'm probably missing something there....

jayvdb:
> I would guess that trying to figure out whether a given identifier needs to be quoted or not will be slower than just always quoting

I guess so as well.

The only reason I can give is that unnecessary quotes is harder for humans to read.

Two places where this turns up is

1. the raw up/down SQL files.  If they are being reviewed by a human, having quotes appear only on reserved words gives reviewers a hint that the name of those columns are likely to be a source of bugs in the future, and maybe it is better to rename the struct member name now before the change is approved.  Avoiding tech debt.
2. error logs are easier to read if there isnt a lot of unnecessary quoting.  The error handler often will be using `{:?}` on objects within the error message, which adds a layer of quotes  as seen at https://github.com/Electron100/butane/issues/102, and often the error handler will also be doing `blah blah: \"{:?}\"` adds another set of quotes for the brain to contend with.

(1) can be kept by only using `quote_reserved_word()` in DDL.  The overhead shouldn't be a problem for offline once-off activities like generating the migrations.

If the runtime performance hit is a problem, then I would prefer to put the extra dependency & `quote_reserved_word()` behind a feature, so butane always quotes if the feature is not enabled.

jayvdb:
We can also copy `sqlparser::keywords::ALL_KEYWORDS` into butane.  The description at https://docs.rs/sqlparser/latest/sqlparser/keywords/index.html suggests this list is too broad and should be trimmed.

Electron100:
Keeping the up/down sql files more readable is fair. 
I realized belatedly after trying it out, that always quoting the runtime operations while not quoting the sql files (unless necessary) doesn't work due to the case normalization of unquoted identifiers, as mentioned in section 4.1.1 here https://www.postgresql.org/docs/current/sql-syntax-lexical.html](url).

I don't love the performance hit of the linear array search here, but since this fixes an issue with correctness, I'm going to go ahead and merge it as-is then try to speed it up in a following change.

jayvdb:
ah, we should probably capitalise the words that are being quoted.

Electron100:
The Postgres documentation notes that it folds unquoted identifiers to lower-case rather than upper-case as specified by the standard. So if we were to anticipate the case folding rules we'd need to make more of the sql logic db-specific.

PR #108, opened 2023-06-12
Add core tests for uuid support

jayvdb:


butane_core/src/lib.rs
```diff
@@ -6,6 +6,10 @@ use std::cmp::{Eq, PartialEq};
 use std::default::Default;
 use thiserror::Error as ThisError;
 
+#[cfg(test)]
+#[macro_use]
+extern crate assert_matches;
```
Electron100, line 11: Why is the `#[macro_use]`/`extern crate` necessary? I thought that went away with 2018. I pulled your branch locally and deleted this block and the tests still seemed to compile and run...

jayvdb: I was just following their README.  I'll remove and then merge it later today, unless you see anything else that can be improved..?

Issue #109, opened 2023-06-14
save() silently disregards unsaved Many relationships 


jayvdb:
If we remove the `tag.save(..)` at https://github.com/Electron100/butane/blob/master/butane/tests/common/blog.rs#L81 , the `many.rs` tests fail, but only fail at the assertions on the number of tags after the posts are re-loaded from the database.

The `Post.save(..)` should detect that the related tags are not saved, and either save them, or error.
IMO doing the save is the most desirable, and isnt a breaking change as anyone who had code that triggers the related save's had broken code.

jayvdb:
Due to https://github.com/Electron100/butane/pull/158 , postgres now catches this, resulting in
```
called `Result::unwrap()` on an `Err` value: Postgres(Error { kind: Db, cause: Some(DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState(E23503), message: "insert or update on table \"bar\" violates foreign key constraint \"bar_foo_fkey\"", detail: Some("Key (foo)=(1) is not present in table \"foo\"."), hint: None, position: None, where_: None, schema: Some("public"), table: Some("bar"), column: None, datatype: None, constraint: Some("bar_foo_fkey"), file: Some("ri_triggers.c"), line: Some(2608), routine: Some("ri_ReportViolation") }) })
```
sqlite ignores it.

Issue #112, opened 2023-06-15
Add upsert support


jayvdb:
In a multi-threaded / async program, the order of events is not predictable, so it is safer to able to use `upsert` instead of `insert` or `update`.

e.g. in the `blog` example, if there are two threads and they both want to create a tag "danger", this will result in an unique constraint violation in the `insert` operation for one of them.  Doing a query for the tag "danger" to determine whether to insert can help, but the query+insert it isnt atomic so another thread could have created the tag between the query and the insert operations .

PR #113, opened 2023-06-15
WIP: Auto-save Many new_values items

jayvdb:
Attempt at solving https://github.com/Electron100/butane/issues/109
It still has lots of todo's:

- This is a very basic implementation, which will break if the schema has circular `Many` references. I'm not sure if butane can support this at the moment. (There is self-referential fkey support)  If so, this auto-save feature will need to be feature flagged.
- Cloning the items can result in lots of memory being used for large objects - I have another implementation which avoids this, but it isnt working yet
- ...

### DIFF
--- a/butane_core/src/many.rs
+++ b/butane_core/src/many.rs
@@ -24,13 +24,13 @@ fn default_oc<T>() -> OnceCell<Vec<T>> {
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct Many<T>
 where
-    T: DataObject,
+    T: DataObject + Clone,
 {
     item_table: Cow<'static, str>,
     owner: Option<SqlVal>,
     owner_type: SqlType,
     #[serde(skip)]
-    new_values: Vec<SqlVal>,
+    new_values: Vec<T>,
     #[serde(skip)]
     removed_values: Vec<SqlVal>,
     #[serde(skip)]
@@ -39,7 +39,7 @@ where
 }
 impl<T> Many<T>
 where
-    T: DataObject,
+    T: DataObject + Clone,
 {
     /// Constructs a new Many. `init` must be called before it can be
     /// loaded or saved (or those methods will return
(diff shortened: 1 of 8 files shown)


butane_core/src/many.rs
```diff
@@ -108,13 +111,14 @@ where
     pub fn save(&mut self, conn: &impl ConnectionMethods) -> Result<()> {
         let owner = self.owner.as_ref().ok_or(Error::NotInitialized)?;
         while !self.new_values.is_empty() {
+            let mut new_value = self.new_values.pop().unwrap();
+            eprintln!("saving item in {:?}", self.item_table);
+            // ignore failures, as they will be unique constraints
```
jayvdb, line 116: or use https://github.com/Electron100/butane/issues/112

Electron100:
I like the goal, but I'm not wild about the clones (or of the fact that it's a breaking change by requiring the type be clonable), or of the fact that it tries to save again unconditionally even when it may not be necessary. I posted #116 for comparative discussion. It's much less ambitious in that it doesn't achieve the full goal of saving automatically, but it at least makes it harder to accidentally shoot oneself in the foot.

If we want to do automatic save, I'd lean more towards doing that save right away in the `add` method rather than cloning to do it later in the `save`. Of course that would require adding the connection as a parameter, which would be a breaking change too.
As a side tangent, while I like a lot of the decisions in Rust, the biggest thing I miss compared to the languages I use in my day job (primarily Kotlin with smatterings of Java, C++, and C#) is method overloading. I slightly-grudgingly understand the reasoning for not supporting type-based overloading, but I do wish we could at least have overloading on arity.

jayvdb:
fyi, now `.saved` is gone, I will re-work this PR to create a separate method to recursively/deeply save.

jayvdb:
This wont be possible due to `Many.add` now checking that the added item is already saved.

PR #116, opened 2023-06-17
Improve checking for unitialized values added to a Many

Electron100:
This is a potential partial fix for #109 in that it reduces the risk of accidentally adding to a many an object with a primay key that isn't actually recorded in the database yet.

It does not provide an ideal solution in that it merely detects, rather than automatically saving.

Posting for comparison and discussion with #113


jayvdb (approved):

jayvdb:
I expect https://github.com/Electron100/butane/pull/113 will be feature flagged when it is finished, so this PR is needed even if that PR is merged.  We need this PR as the default.  Also need the same for fkeys.

jayvdb:
fwiw, I have nearly finished doing the same for fkey

PR #118, opened 2023-06-19
WIP: Auto-save ForeignKey values

jayvdb:


jayvdb:
fyi, now `.saved` is gone, I will re-work this PR to create a separate method to recursively/deeply save.

PR #130, opened 2023-07-17
Add cli --path argument

jayvdb:


butane_cli/src/main.rs
```diff
@@ -1,18 +1,26 @@
...
+    }
     let app = clap::Command::new("butane")
         .version(env!("CARGO_PKG_VERSION"))
         .author("James Oakley <james@electronstudio.org>")
         .about("Manages butane database migrations")
         .max_term_width(80)
+        .arg(
+            Arg::new("path").short('p').long("path")
+            .default_value(WORKING_DIR_PATH.as_os_str())
+            .value_parser(value_parser!(PathBuf))
```
Electron100, line 22: Could you add help text for this? Something like "select directory to look for butane state in"

Issue #122, opened 2023-06-27
Stable ordering for Many.load


jayvdb:
The `Iterator` returned by `Many.load` doesn't have any default sorting.

It would make sense to use the primary key as the default sorting for the database query, however there is also the aspect that `.load` also includes `.new_values`, which would need to be manually sorted into the db values.

Also I suspect there is a problem with `.load()` that it doesnt check/flush `.removed_values` before reading from the database.

jayvdb:
Or perhaps we should have a `Many.query() -> Query` that creates a Query object pre-filtered to the `Many`'s objects, so then sorting can be specified before the query is executed.

Electron100:
I like that second idea. I don't see any reason that the output of `Many.load` needs to *always* be sorted, but I agree that the *ability* to sort it would be valuable.

PR #131, opened 2023-07-17
Add Many.query

jayvdb:
Closes https://github.com/Electron100/butane/issues/122

### DIFF
--- a/butane_core/src/many.rs
+++ b/butane_core/src/many.rs
@@ -1,7 +1,7 @@
 //! Implementation of many-to-many relationships between models.
 #![deny(missing_docs)]
 use crate::db::{Column, ConnectionMethods};
-use crate::query::{BoolExpr, Expr, Query, OrderDirection};
+use crate::query::{BoolExpr, Expr, OrderDirection, Query};
 use crate::{DataObject, Error, FieldType, Result, SqlType, SqlVal, ToSql};
 use once_cell::unsync::OnceCell;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


jayvdb:
While this allows doing custom querys, it doesnt provide a way to put the query results back into the `Many`. Still thinking about that.  Adding "load from query results" is simple in principle, however it will mean there is a way to bypass the usual loading mechanism.

butane_core/src/many.rs
```diff
@@ -125,23 +125,30 @@ where
...
+            col: T::PKCOL,
+            tbl2: self.item_table.clone(),
+            tbl2_col: "has",
+            expr: Box::new(BoolExpr::Eq("owner", Expr::Val(owner.clone()))),
+        }))
+    }
+
+    /// Loads the values referred to by this many relationship from a
+    /// database query if necessary and returns a reference to them.
+    pub fn load_query(
```
Electron100, line 145: The `query`/`load_query` methods have a couple rough edges
1. The returned query already has the filter set. Any attempts to set the filter will result in a query that's no longer actually associated with the `Many`.
2. `load_query` assumes that they query provided was returned from `query` and that it does not do any further filtering of the results (if it did, always adding in the unsaved items would be invalid since they might not match the filter).

I don't think it's necessary to remove these rough edges for now until there's actually a clear use case for these methods being more flexible, but it would be nice to document the expectations in the method comments. Alternatively these two methods could stay private for now, with `load_ordered` the only new public method.

Not blocking.

Electron100 (approved):

PR #139, opened 2023-07-29
Use clio to check base directory

jayvdb:


jayvdb:
I am not convinced this is worth the extra dependency.
the error looks like
> error: Invalid value for --path <path>: Invalid path "foo": No such file or directory (os error 2)

PR #140, opened 2023-07-31
Use line length of 100

jayvdb:
These lines were not being reformatted.
To find them, nightly rustfmt is needed using https://github.com/rust-lang/rustfmt/issues/3391

Is this worth the blame changes?
If so, should I also add CI to install nightly rustfmt and check with that option?


Electron100:
I'm fine with the blame changes.
Putting nightly rustfmt with `error_on_line_overflow` in CI seems nice-to-have, but hardly mission critical.

Electron100 (approved):

PR #141, opened 2023-07-31
Group imports: std, external, crate

jayvdb:
This is what we get with rustfmt nightly option group_imports = "StdExternalCrate".

Related to https://github.com/Electron100/butane/pull/140

Happy to abandon if you dont think it is helpful.


Electron100:
The actual specifics of how imports are grouped lies firmly in the set of things I have no real strong opinion on. But I don't really like this being a nightly-only option. I don't want to force contributors to use rustfmt nightly, and if some people use the option and others don't then we'll see rustfmt stomping on the import area for unrelated changes.

Theoretically one could raise similar concerns on #140, but I'm less concerned about the use of nightly there because it's an edge-case for detecting sections not formatted due to bugs in rustfmt, and should be much more rare than differences in import organization.

If this option gets standardized then I'd be fine with enabling it through `.rustfmt.toml`

jayvdb:
> if some people use the option and others don't then we'll see rustfmt stomping on the import area for unrelated changes.

Coming back to this, if the imports are sorted using this nightly option, and then someone uses stable without this option, stable will not re-arrange the imports.  That is because this feature splits the groups with a blank line, and the stable formatter only rearranges imports only within any existing groups separated by blank lines.

Electron100:
Ah, thanks for that clarification. In that case I have no objection to this moving forward.

PR #142, opened 2023-07-31
WIP: Fake relationship members

jayvdb:
This is a demo of 
- adding a fake value into a `ForeignKey<T>` looks doable, but requires `T: Clone` in order for the fake `T` to be stored in the db
- why adding fake values into a `Many<T>` looks to be incompatible with how `Many` is designed

I think it is worthwhile to generate a fake value for `ForeignKey<T>` , if only because faking with `Self::new_raw()` is really bad, because it exposes an invalid state that can not otherwise easily be reached.

### DIFF
--- a/butane_core/src/fkey.rs
+++ b/butane_core/src/fkey.rs
@@ -8,7 +8,7 @@ use std::borrow::Cow;
 use std::fmt::{Debug, Formatter};
 
 #[cfg(feature = "fake")]
-use fake::{Dummy, Faker};
+use fake::{Dummy, Faker, Fake};
 
 /// Used to implement a relationship between models.
 ///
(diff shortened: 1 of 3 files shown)


butane_core/src/many.rs
```diff
@@ -215,9 +215,15 @@ impl<T: DataObject> Default for Many<T> {
...
+/// Fake data support generates one value,
+/// however it is impossible to the many relationship to it.
+/// If we put them in `new_values`, we can only store the `.pk()`
+/// which makes it impossible to save the value being related to.
+impl<T: DataObject + Dummy<Faker>> Dummy<Faker> for Many<T> {
     fn dummy_with_rng<R: rand::Rng + ?Sized>(_: &Faker, _rng: &mut R) -> Self {
-        Self::new()
+        let obj = Faker.fake::<T>();
+        let ret = Self::new();
+        ret.all_values.set(vec![obj]).ok();
```
jayvdb, line 226: Values in `all_values` are presumed to be already saved, and thus not savable.

The only way I can think to get this working is: a `pub fn` which loads from the db, and compares it with all parts of this `Many` to determine which DML is needed to "sync" to the db.  This could be of value if there are multiple threads, and the current thread wants to make its version of the `Many<T>` to be authoritative, despite changes that other threads may have made to the same `Many<T>`.   There are several minefields down that path.

butane_core/src/fkey.rs
```diff
@@ -206,9 +206,14 @@ where
 }
 
 #[cfg(feature = "fake")]
-/// Fake data support is currently limited to empty ForeignKey relationships.
-impl<T: DataObject> Dummy<Faker> for ForeignKey<T> {
-    fn dummy_with_rng<R: rand::Rng + ?Sized>(_: &Faker, _rng: &mut R) -> Self {
-        Self::new_raw()
+/// A dummy value for T will be created, however it also has to have the Clone
+/// trait in order for the mutable value to be obtained so that it can be saved.
+impl<T: DataObject + Dummy<Faker>> Dummy<Faker> for ForeignKey<T> {
```
jayvdb, line 211: If we proceed with this, and `.clone()` is needed to get the `T` to be saved, IMO we should do:
```diff
-impl<T: DataObject + Dummy<Faker>>...
+impl<T: Clone + DataObject + Dummy<Faker>>...
``` 
and then document how to save the fake data.

PR #145, opened 2023-09-14
Update getting-started.md

jayvdb:


docs/getting-started.md
```diff
@@ -166,7 +166,7 @@ Then we can use them in our `lib.rs`:
 ```rust
 pub mod models;
 
-use models::{Blog, Post};
+pub use models::{Blog, Post};
```
jayvdb, line 169: this is added so that if someone does a build at this stage, they don't get warnings.

PR #148, opened 2023-10-22
Allow running tests with various features

jayvdb:


butane/tests/common/blog.rs
```diff
@@ -36,13 +38,29 @@ pub struct Post {
     pub tags: Many<Tag>,
     pub blog: ForeignKey<Blog>,
 }
+
+#[cfg(not(feature = "datetime"))]
+#[model]
+#[derive(Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "fake", derive(Dummy))]
+pub struct Post {
```
Electron100, line 46: Duplicating the whole struct should not have been necessary, but after some poking I see why it is. Filed #150 to track

Electron100 (approved):

PR #151, opened 2023-10-23
Remove ObjectState (breaking change)

Electron100:
ObjectState never really took on as large a role I had originally envisioned, while at the same time being quite clunky -- an extra field automatically injected into all model types. It had a single member `saved`, which tracks whether the object had previously been saved or not. Most of the time this doesn't matter though, especially since #117 changed save to upsert. I believe the only remaining case where it does matter is for automatic creation of primary keys, and those were operating in a semantic grey area where creating an unitialized primary key required setting it to `-1` or similar.

This change does the following
1. Removes `ObjectState` and the auto-created `state` member for each model
2. Adds a new `AutoPk<T>` type which is used as the type for fields which represent automatic primary keys. This tracks whether the primary key has already been generated or not.

### DIFF
--- a/butane/tests/basic.rs
+++ b/butane/tests/basic.rs
@@ -224,10 +224,8 @@ testall!(foreign_key);
 fn auto_pk(conn: Connection) {
     let mut baz1 = Baz::new("baz1");
     baz1.save(&conn).unwrap();
-    eprintln!("{:?}", baz1.id);
     let mut baz2 = Baz::new("baz2");
     baz2.save(&conn).unwrap();
-    eprintln!("{:?}", baz2.id);
     let mut baz3 = Baz::new("baz3");
     baz3.save(&conn).unwrap();
     assert!(baz1.id < baz2.id);
--- a/butane/tests/basic.rs
+++ b/butane/tests/basic.rs
@@ -128,7 +128,7 @@ fn basic_crud(conn: Connection) {
 
     // delete
     assert!(foo3.delete(&conn).is_ok());
-    if let Some(butane::Error::NoSuchObject) = Foo::get(&conn, 1).err() {
+    if matches!(Foo::get(&conn, 1).err(), Some(butane::Error::NoSuchObject)) {
     } else {
         panic!("Expected NoSuchObject");
     }
--- a/butane/tests/basic.rs
+++ b/butane/tests/basic.rs
@@ -1,7 +1,6 @@
 use butane::db::Connection;
-use butane::{butane_type, find, model, query};
+use butane::{butane_type, find, model, query, AutoPk, ForeignKey};
 use butane::{colname, prelude::*};
-use butane::{ForeignKey, ObjectState};
 use chrono::{naive::NaiveDateTime, offset::Utc, DateTime};
 use serde::Serialize;
 
@@ -29,7 +28,6 @@ impl Foo {
             bar: 0,
             baz: String::new(),
             blobbity: Vec::new(),
-            state: ObjectState::default(),
         }
     }
 }
@@ -46,23 +44,20 @@ impl Bar {
         Bar {
             name: name.to_string(),
             foo: foo.into(),
-            state: ObjectState::default(),
         }
     }
 }
 
 #[model]
 struct Baz {
-    #[auto]
-    id: i64,
+    id: AutoPk<i64>,
     text: String,
 }
 impl Baz {
     fn new(text: &str) -> Self {
         Baz {
-            id: -1, // will be set automatically when saved
+            id: AutoPk::default(),
             text: text.to_string(),
-            state: ObjectState::default(),
         }
     }
 }
@@ -73,10 +68,7 @@ struct HasOnlyPk {
 }
 impl HasOnlyPk {
     fn new(id: i64) -> Self {
-        HasOnlyPk {
-            id,
-            state: ObjectState::default(),
-        }
+        HasOnlyPk { id }
     }
 }
 
@@ -91,7 +83,6 @@ impl SelfReferential {
         SelfReferential {
             id,
             reference: None,
-            state: ObjectState::default(),
         }
     }
 }
@@ -229,8 +220,10 @@ testall!(foreign_key);
 fn auto_pk(conn: Connection) {
     let mut baz1 = Baz::new("baz1");
     baz1.save(&conn).unwrap();
+    eprintln!("{:?}", baz1.id);
     let mut baz2 = Baz::new("baz2");
     baz2.save(&conn).unwrap();
+    eprintln!("{:?}", baz2.id);
     let mut baz3 = Baz::new("baz3");
     baz3.save(&conn).unwrap();
     assert!(baz1.id < baz2.id);
@@ -344,7 +337,6 @@ fn basic_time(conn: Connection) {
         naive: now.naive_utc(),
         utc: now,
         when: now,
-        state: ObjectState::default(),
     };
     time.save(&conn).unwrap();
 
--- a/butane/tests/basic.rs
+++ b/butane/tests/basic.rs
@@ -1,6 +1,9 @@
+#![allow(clippy::disallowed_names)]
+
 use butane::db::Connection;
 use butane::{butane_type, find, model, query, AutoPk, ForeignKey};
 use butane::{colname, prelude::*};
+#[cfg(feature = "datetime")]
 use chrono::{naive::NaiveDateTime, offset::Utc, DateTime};
 use serde::Serialize;
 
@@ -87,6 +90,7 @@ impl SelfReferential {
     }
 }
 
+#[cfg(feature = "datetime")]
 #[model]
 #[derive(Debug, Default, PartialEq, Clone)]
 struct TimeHolder {
@@ -330,6 +334,7 @@ fn fkey_same_type(conn: Connection) {
 }
 testall!(fkey_same_type);
 
+#[cfg(feature = "datetime")]
 fn basic_time(conn: Connection) {
     let now = Utc::now();
     let mut time = TimeHolder {
@@ -345,6 +350,7 @@ fn basic_time(conn: Connection) {
     // lose some precision when we go to the database.
     assert_eq!(time.naive.timestamp(), time2.naive.timestamp());
 }
+#[cfg(feature = "datetime")]
 testall!(basic_time);
 
 fn basic_load_first(conn: Connection) {
--- a/butane_codegen/src/lib.rs
+++ b/butane_codegen/src/lib.rs
@@ -26,10 +26,6 @@ mod filter;
 /// ## Helper Attributes
 /// * `#[table = "NAME"]` used on the struct to specify the name of the table (defaults to struct name)
 /// * `#[pk]` on a field to specify that it is the primary key.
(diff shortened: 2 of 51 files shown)


butane_core/src/codegen/dbobj.rs
```diff
@@ -29,12 +29,40 @@ pub fn impl_dbobject(ast_struct: &ItemStruct, config: &Config) -> TokenStream2 {
...
-    let mut post_insert: Vec<TokenStream2> = Vec::new();
-    add_post_insert_for_auto(&pk_field, &mut post_insert);
-    post_insert.push(quote!(self.state.saved = true;));
+    let save_core = if is_auto(&pk_field) {
+        let pkident = pk_field.ident.clone().unwrap();
+        let values_no_pk: Vec<TokenStream2> = push_values(ast_struct, |f: &Field| f != &pk_field);
+        let save_cols = columns(ast_struct, |f| !is_auto(f) && f != &pk_field);
+        quote!(
+            // Since we expect our pk field to be invalid and to be created by the insert,
+            // we do a pure insert or update, no upsert allowed.
```
jayvdb, line 41: not particular relevant at this stage, but might be worth using to clarify in this comment, but I believe we can do upserts with auto-increment (at least pg, mysql, havent checked sqlite), however this requires unique key constraints on other fields to tell the db engine when to update instead of inserting, and we dont have a way to do unique constraints yet.

Electron100: Good point

butane_core/src/autopk.rs
jayvdb, line 1: Please add docs to this new file

butane_core/src/autopk.rs
```diff
@@ -0,0 +1,109 @@
+use super::{FieldType, FromSql, PrimaryKeyType, Result, SqlType, SqlVal, SqlValRef, ToSql};
```
jayvdb, line 1: optional, but IMO worth considering moving towards https://github.com/Electron100/butane/pull/141 - I've added a comment there to clarify how the two would interact with each other, in case it is relevant.

```diff
diff --git a/butane_core/src/autopk.rs b/butane_core/src/autopk.rs
index 16f3888..9c7d482 100644
--- a/butane_core/src/autopk.rs
+++ b/butane_core/src/autopk.rs
@@ -1,7 +1,9 @@
-use super::{FieldType, FromSql, PrimaryKeyType, Result, SqlType, SqlVal, SqlValRef, ToSql};
-use serde::{Deserialize, Serialize};
 use std::cmp::{Ordering, PartialOrd};
 
+use serde::{Deserialize, Serialize};
+
+use super::{FieldType, FromSql, PrimaryKeyType, Result, SqlType, SqlVal, SqlValRef, ToSql};
+
 #[derive(Clone, Debug, Default, Serialize, Deserialize)]
 pub struct AutoPk<T: PrimaryKeyType> {
     inner: Option<T>,
```

examples/getting_started/src/models.rs
```diff
@@ -1,11 +1,11 @@
...
-use butane::{model, ForeignKey, Many, ObjectState};
+use butane::AutoPk;
+use butane::{model, ForeignKey, Many};
 
 #[model]
 #[derive(Debug, Default)]
 pub struct Blog {
-    #[auto]
-    pub id: i64,
+    pub id: AutoPk<i64>,
```
jayvdb, line 8: docs/getting_started.md needs updating.

butane_core/src/codegen/dbobj.rs
```diff
@@ -29,12 +29,43 @@ pub fn impl_dbobject(ast_struct: &ItemStruct, config: &Config) -> TokenStream2 {
...
-    let mut post_insert: Vec<TokenStream2> = Vec::new();
-    add_post_insert_for_auto(&pk_field, &mut post_insert);
-    post_insert.push(quote!(self.state.saved = true;));
+    let save_core = if is_auto(&pk_field) {
+        let pkident = pk_field.ident.clone().unwrap();
+        let values_no_pk: Vec<TokenStream2> = push_values(ast_struct, |f: &Field| f != &pk_field);
+        let save_cols = columns(ast_struct, |f| !is_auto(f) && f != &pk_field);
+        quote!(
+            // Since we expect our pk field to be invalid and to be created by the insert,
+            // we do a pure insert or update, no upsert allowed. Note that some database backends
```
jayvdb, line 41: Why cant we use upset for updates?  It seems like we could.

butane_codegen/src/lib.rs
```diff
@@ -42,9 +38,8 @@ mod filter;
 /// #[model]
 /// #[table = "posts"]
 /// pub struct Post {
-///   #[auto]
```
jayvdb, line 45: we could re-implement `#[auto]` so that it injects the new `AutoPk`.
That would remove the largest part that is 'breaking'.

We could also update `#[model]` so that it removes any `state` field automatically, but I guess that isnt worth doing as the code would still fail to compile due to the `use butane::ObjectState` which would be in the code.  Maybe have a dummy `ObjectState` type.  There are deprecation features of Rust we could use to advise people to remove that `use`.

jayvdb: fyi, I am happy to attempt this after your PR, if you agree with it.

examples/getting_started/src/models.rs
jayvdb: I can do this in a followup PR, if you like

jayvdb: oh its done.

butane_core/src/lib.rs
```diff
@@ -29,31 +28,28 @@ pub use sqlval::*;
...
-/// Used internally by butane to track state about the object.
-///
-/// Includes information such as whether it has actually been created
-/// in the database yet. Butane automatically creates the field
-/// `state: ObjectState` on `#[model]` structs. When initializing the
-/// state field, use `ObjectState::default()`.
-#[derive(Clone, Debug, Default, Ord, PartialOrd)]
-pub struct ObjectState {
-    pub saved: bool,
+pub trait ToPkSql<T>
```
jayvdb, line 31: docs missing

butane_core/src/autopk.rs
```diff
@@ -0,0 +1,118 @@
+//! Contains the [AutoPk] type for autoincrementing primary keys.
+
+use super::{FieldType, FromSql, PrimaryKeyType, Result, SqlType, SqlVal, SqlValRef, ToSql};
+use serde::{Deserialize, Serialize};
+use std::cmp::{Ordering, PartialOrd};
+
+/// Wrapper around a `PrimaryKeyType` to indicate the the primary key
```
jayvdb, line 7: wrap `PrimaryKeyType` in square brackets?

the the => that the

butane_core/src/autopk.rs
```diff
@@ -0,0 +1,118 @@
...
+        } else {
+            self.inner.eq(&other.inner)
+        }
+    }
+}
+
+impl<T: PrimaryKeyType> FieldType for AutoPk<T> {
+    const SQLTYPE: SqlType = T::SQLTYPE;
+    /// Reference type. Used for ergonomics with String (which has
+    /// reference type str). For most, it is Self
```
jayvdb, line 85: trailing `.`?

butane_core/src/codegen/mod.rs
```diff
@@ -297,7 +258,7 @@ fn pk_field(ast_struct: &ItemStruct) -> Option<Field> {
 }
 
 fn is_auto(field: &Field) -> bool {
-    field.attrs.iter().any(|attr| attr.path().is_ident("auto"))
+    get_foreign_type_argument(&field.ty, &AUTOPK_TYNAMES).is_some()
```
jayvdb, line 261: "get_foreign_type_argument" probably needs a more generic name..?

butane_core/src/codegen/dbobj.rs
Electron100: Well we could, but at that point we know the object has already been created so there's no real point. We're in a separate `quote!` block from the the upsert so I don't think we can merge easily. Let me know if you see something more elegant though.
I'll reword slightly to try to be more clear.

jayvdb: Ya, it isnt beneficial to use upsert. "no upsert allowed" seemed a bit too strongly asserting it would be a problem.
Rewording the comment is sufficient.

butane_core/src/lib.rs
Electron100: Thanks for calling that out. My memory is going, and I can't seem to remember why I created this extra trait. Things seem to work without it, so I'm going to remove it...

butane_codegen/src/lib.rs
Electron100: Good suggestion. Let me try to see if I can find time this weekend to do that.

butane_core/src/autopk.rs
Electron100: Thanks for the update there. I won't take that change *in this PR* but am OK with it going forward in #141. Let's continue the discussion there.

butane/tests/basic.rs
```diff
@@ -233,8 +224,10 @@ testall!(foreign_key);
 fn auto_pk(conn: Connection) {
     let mut baz1 = Baz::new("baz1");
     baz1.save(&conn).unwrap();
+    eprintln!("{:?}", baz1.id);
```
jayvdb, line 227: are these two eprintln still useful?

butane_core/src/autopk.rs
```diff
@@ -0,0 +1,118 @@
...
+    {
+        Ok(AutoPk::with_value(T::from_sql(val)?))
+    }
+}
+
+impl<T: PrimaryKeyType> ToSql for AutoPk<T> {
+    fn to_sql(&self) -> SqlVal {
+        self.inner
+            .as_ref()
+            .expect("PK is not generated yet!")
```
jayvdb, line 58: DRY this message? create a private func for 
```rust
.as_ref().expect("PK is not generated yet!")
```

butane_core/src/autopk.rs
```diff
@@ -0,0 +1,118 @@
...
+            None => false,
+        }
+    }
+}
+
+impl<T: PrimaryKeyType + std::fmt::Display> std::fmt::Display for AutoPk<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::result::Result<(), std::fmt::Error> {
+        match &self.inner {
+            Some(val) => val.fmt(f),
+            None => write!(f, "null"),
```
jayvdb, line 101: why pretend to be valid SQL here?
"not_initialised" or similar says the same without looking valid

jayvdb (approved):
a few comments added, but overall this is awesome and lets get it in asap

butane_core/src/autopk.rs
Electron100: Fair, updated

butane_core/src/autopk.rs
Electron100: Updated

butane/tests/basic.rs
Electron100: No, thanks for noticing

PR #153, opened 2023-10-24
Lint docs

jayvdb:
This fixes all the warnings like

```
error: redundant explicit link target
   --> butane_core/src/lib.rs:222:25
    |
222 | /// See also [`SqlVal`][crate::SqlVal].
    |               --------  ^^^^^^^^^^^^^ explicit target is redundant
    |               |
    |               because label contains path that resolves to same destination
    |
    = note: when a link's destination is not specified,
            the label is used to resolve intra-doc links
help: remove explicit link target
    |
222 | /// See also [`SqlVal`].
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,32 +1,38 @@
+CARGO := rustup run stable cargo
+CARGO_NIGHTLY := $(subst stable, nightly, $(CARGO))
 
 all : build
 
 build :
-	cargo check
+	$(CARGO) check
 	# build some intermediate configuration to test different feature combinations
-	cd butane && cargo check --features pg
-	cd butane && cargo check --features pg,datetime
-	cd butane && cargo check --features sqlite
+	cd butane && $(CARGO) check --features pg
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1,32 +1,38 @@
+CARGO := rustup run stable cargo
```
Electron100, line 1: Can you say more about why this change is necessary?

jayvdb: Its not strictly necessary.  My `cargo` is the distro provided one, which is pinned, and I prefer to keep it like that because I do a lot of distro packaging.  `rustup` is how I get a real unpinned stable/nightly/etc.

jayvdb: If you dont want this, I can change it to
```
CARGO := cargo +stable
```
I wrote `CARGO_NIGHTLY` so that it will work with ^ syntax and the rustup syntax.

Then I can change one line, and the rest will work for me.

Electron100 (approved):

Issue #21, opened 2022-12-24
Database schema generation seems to be broken


MrJico:
First, I would like to thank you for this beautifull 😍project😍.

so i oppened this issue becuase i have tested this crate on two diffrent setups and both with backends(sqllite, pg). but the generated sql code for the relations is ether completly broken or there is something wrong with my setup.

basicly i copy pasted the getting started page and made the migrations. which in theory should generate all the necessary sql code for the relation ships. but it does not.

here is an example of the sql output: ⬇️

```sql
CREATE TABLE Post (
id SERIAL NOT NULL PRIMARY KEY,
title TEXT NOT NULL,
body TEXT NOT NULL,
published BOOLEAN NOT NULL,
blog BIGINT NOT NULL,
byline TEXT 
);
CREATE TABLE Tag (
tag TEXT NOT NULL PRIMARY KEY
);
CREATE TABLE Blog (
id BIGSERIAL NOT NULL PRIMARY KEY,
name TEXT NOT NULL
);
CREATE TABLE Post_tags_Many (
owner INTEGER NOT NULL,
has TEXT NOT NULL
);
CREATE TABLE IF NOT EXISTS butane_migrations (
name TEXT NOT NULL PRIMARY KEY
);
```

and here in this image visualized with dbeaver:
![image](https://user-images.githubusercontent.com/63163262/209447812-57965033-7c64-4556-b62e-c56598c726ff.png)

yet there are not relations(foreign keys, reffrences) generated for the tables.

here an example of what the generated sql code should be: ⬇️

```sql
CREATE TABLE Blog (
    id BIGSERIAL NOT NULL PRIMARY KEY,
    name TEXT NOT NULL
);

CREATE TABLE Post (
    id SERIAL NOT NULL PRIMARY KEY,
    title TEXT NOT NULL,
    body TEXT NOT NULL,
    published BOOLEAN NOT NULL,
    blog INTEGER REFERENCES Blog(id),
    byline TEXT
);

CREATE TABLE TAG (
    tag TEXT NOT NULL PRIMARY KEY
);

CREATE TABLE Post_tags_Many (
    owner SERIAL REFERENCES Post(id),
    has TEXT REFERENCES Tag(tag),
    PRIMARY KEY (owner, has)
);


CREATE TABLE IF NOT EXISTS butane_migrations (
name TEXT NOT NULL PRIMARY KEY
);
```
and in this image visualized with dbeaver:

[...]

PR #155, opened 2023-11-01
Force stable ordering of migration operations

jayvdb:
Preliminary work related to https://github.com/Electron100/butane/issues/21

The order of tables becomes important when the SQL for each table needs to refer to other tables.
Especially for sqlite, where constraints need to be in the `ADD TABLE` and cant be added using `ALTER TABLE` c.f. https://www.sqlite.org/omitted.html

### DIFF
--- a/butane_core/src/migrations/adb.rs
+++ b/butane_core/src/migrations/adb.rs
@@ -5,7 +5,7 @@
 use crate::{Error, Result, SqlType, SqlVal};
 use serde::{de::Deserializer, de::Visitor, ser::Serializer, Deserialize, Serialize};
 use std::cmp::Ordering;
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeSet, HashMap};
 
 /// Identifier for a type as used in a database column. Supports both
 /// [`SqlType`] and identifiers known only by name.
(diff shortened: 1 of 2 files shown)


butane_core/src/migrations/adb.rs
```diff
@@ -461,3 +483,24 @@ fn diff_table(old: &ATable, new: &ATable) -> Vec<Operation> {
...
     ops
 }
+
+#[cfg(test)]
+mod tests {
+    use std::collections::HashSet;
+
+    #[test]
+    /// Demostration that .difference() results are not stable.
+    fn show_unstable_migration_operation_order() {
```
jayvdb, line 493: this test can be removed before merge - it doesnt test butane - it only shows using HashSet and difference/intersection is problematic.

I am now going to see whether using BTreeSet or `indexmap` crate provides another way of stablising the order without needing to alpha sort.

jayvdb: Using BTreeSet doesnt offer any improvements - only the explicit `.sort()` can be removed.

using crate `indexmap` does allow "insertion order" of new tables and new columns, but it also requires that `ADB.tables` becomes a `indexmap::IndexMap` , and that means we need to enable indexmap feature `serde`.

I don't have a strong preference for insertion order vs alpha order.

db table explorers almost always show table lists and column lists in alpha order, so the order we insert them isnt likely to ever impact anyone. 

The only pro/con I can think of is that alpha order may make resolving merge conflicts a bit easier, if/when we ever get to helping the user resolve conflicting migrations during a rebase.

jayvdb: in case it is useful for reviewing, here is what was here
```rust
        let old = ["a", "b"];
        let new = ["d", "b", "c"];
        let new_names: HashSet<&str> = new.iter().copied().collect();
        let old_names: HashSet<&str> = old.iter().copied().collect();
        let added_tables = new_names.difference(&old_names);
        let mut added_tables: Vec<&str> = added_tables.copied().collect();

        // Remove this line, and running this multiple times will result in a mix of passes and failures.
        added_tables.sort();

        assert_eq!(added_tables, vec!["c", "d"]);
```

Electron100 (approved):

PR #157, opened 2023-11-03
Add migration unit tests

jayvdb:
Creates a pub "create_many_table" so that it is easier to build unit test cases outside of `src/` . Happy to move the tests into test submodules if you prefer that.

### DIFF
--- a/butane_core/tests/migration.rs
+++ b/butane_core/tests/migration.rs
@@ -137,3 +137,275 @@ fn stable_add_column_alpha_order() {
         ]
     );
 }
+
+#[test]
+fn add_table_fkey() {
+    let known_int_type = DeferredSqlType::KnownId(TypeIdentifier::Ty(SqlType::Int));
+
+    let old = ADB::default();
+    let mut new = ADB::default();
+    let mut table_a = ATable::new("a".to_owned());
+    let column = AColumn::new(
+        "id".to_owned(),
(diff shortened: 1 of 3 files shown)


butane_core/tests/migration.rs
```diff
@@ -137,3 +137,484 @@ fn stable_add_column_alpha_order() {
...
+
+    assert_eq!(
+        ops,
+        vec![
+            Operation::AddTable(table_a),
+            Operation::AddTable(resolved_table_b.clone()),
+        ]
+    );
+
+    // The above is a duplicate of test "add_renamed_table_fkey"
```
jayvdb, line 321: need to dedup these tests

butane_core/tests/migration.rs
```diff
@@ -137,3 +137,275 @@ fn stable_add_column_alpha_order() {
...
+    assert_eq!(
+        ops,
+        vec![
+            Operation::AddTable(table_a),
+            Operation::AddTable(resolved_table_b.clone()),
+        ]
+    );
+}
+
+fn create_add_renamed_table_fkey_ops() -> (Vec<Operation>, ADB, ATable, ATable) {
```
Electron100, line 198: Could you add a method comment here just to summarize what this helper is doing?

Electron100 (approved):

PR #158, opened 2023-11-03
Add database constraints

### LINKED ISSUE #21: Database schema generation seems to be broken

jayvdb:
Closes #21

~~Still a few TODOs,~~ ~~including rebasing on top of https://github.com/Electron100/butane/pull/157~~

Note it doesnt add constraints to existing databases, only new ones.


butane_core/src/migrations/adb.rs
```diff
@@ -318,6 +328,45 @@ impl From<TypeIdentifier> for DeferredSqlType {
     }
 }
 
+/// Abstract representation of a database column reference constraint.
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub enum ARef {
+    Literal(ARefLiteral),
+    Deferred(DeferredSqlType),
```
jayvdb, line 335: very likely, this can be simply `Deferred,` without any data, as I suspect that the referred to column is always a available in the column `sqltype`

butane_core/src/migrations/adb.rs
```diff
@@ -318,6 +328,45 @@ impl From<TypeIdentifier> for DeferredSqlType {
...
+}
+
+impl ARef {
+    // not used
```
jayvdb, line 339: TODO: remove, or use?

butane_core/src/migrations/adb.rs
```diff
@@ -318,6 +328,45 @@ impl From<TypeIdentifier> for DeferredSqlType {
...
+
+/// Abstract representation of a database column reference constraint to a literal table and column.
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub struct ARefLiteral {
+    table_name: String,
+    column_name: String,
+}
+
+impl ARefLiteral {
+    pub fn new(table_name: impl Into<String>, column_name: impl Into<String>) -> Self {
```
jayvdb, line 356: add docs for these fns

butane_core/src/migrations/adb.rs
```diff
@@ -388,32 +449,113 @@ impl AColumn {
...
     AddColumn(String, AColumn),
+    /// Remove a table column.
     RemoveColumn(String, String),
+    /// Change a table columns type.
     ChangeColumn(String, AColumn, AColumn),
 }
 
 /// Determine the operations necessary to move the database schema from `old` to `new`.
 pub fn diff(old: &ADB, new: &ADB) -> Vec<Operation> {
+    // TODO: re-order tables so tables without references go first,
```
jayvdb, line 549: need to create a test, effectively same as existing many tests but with a & b swapped to prove this is a problem, and then do the fix

butane_core/src/migrations/adb.rs
jayvdb: IMO this isnt worth entwining this into the `resolve_types` process.

butane_core/src/codegen/migration.rs
```diff
@@ -69,9 +77,20 @@ fn many_table(main_table_name: &str, many_field: &Field, pk_field: &Field) -> AT
         .to_string();
     let many_field_type = get_many_sql_type(many_field)
         .unwrap_or_else(|| panic!("Mis-identified Many field {field_name}"));
+    let pk_field_name = pk_field
+        .ident
+        .clone()
```
Electron100, line 82: This can be `.as_ref` instead of `.clone()`. The `Option` needs to be consumed by expect, but its inner type does not.

butane_core/src/migrations/adb.rs
```diff
@@ -369,6 +430,14 @@ impl AColumn {
     pub fn default(&self) -> &Option<SqlVal> {
         &self.default
     }
+    /// Returns whether this column refers to another column.
+    pub fn reference(&self) -> &Option<ARef> {
+        &self.reference
+    }
+    /// Set another column that this column refers to.
+    pub fn add_reference(&mut self, reference: &ARef) {
```
Electron100, line 438: I don't follow why this method needs to exist. The only usage I see (in `butane_core/codegen/migration.rs`) is at the same time we're constructing an `AColumn`. So why isn't the constructor parameter sufficient?

butane_core/src/migrations/adb.rs
```diff
@@ -318,19 +329,67 @@ impl From<TypeIdentifier> for DeferredSqlType {
...
     unique: bool,
+    /// Default value for the column.
     default: Option<SqlVal>,
+    /// Whether this column refers to another column.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    reference: Option<ARef>,
 }
 impl AColumn {
+    /// Create new column.
+    #[allow(clippy::too_many_arguments)]
```
Electron100, line 392: This should probably be a builder, which would remove the need for the clippy allow and make invocations more readable. Definitely doesn't need to be this PR though, I can try to address it afterwards.

butane_core/src/migrations/adb.rs
jayvdb: Using `AColumn::new_simple` with an optional `::add_reference` is a bit easier to read, but I am happy to switch to using two different invocations of `::new`

Electron100: I'd rather not have the extra method because the column isn't really supposed to be changed after it's created. But I agree with the readability issue, so I won't block here.
The solution to both aspects is to switch to a builder pattern, and I can do that in a follow-up.

Electron100 (approved):

PR #161, opened 2023-11-07
Stabilise embed format

jayvdb:
Currently embed doesnt comply with rustfmt, and the contents of the JSON constantly change order because the data structures use HashMap which has no ordering guarantee.

### DIFF
--- a/butane_cli/src/lib.rs
+++ b/butane_cli/src/lib.rs
@@ -193,16 +193,20 @@ pub fn embed(base_dir: &Path) -> Result<()> {
         copy_migration(&m, &mut new_m)?;
         mem_ms.add_migration(new_m)?;
     }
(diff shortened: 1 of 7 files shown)


butane_cli/src/lib.rs
```diff
@@ -196,13 +196,15 @@ pub fn embed(base_dir: &Path) -> Result<()> {
     let json = serde_json::to_string(&mem_ms)?;
```
jayvdb, line 196: I would also like to change this to to_string_pretty, so that the generated `butane_migrations.rs` is more readable when there are changes to the migrations.

jayvdb:
Created https://github.com/Electron100/butane/issues/162 about an issue that occurred in the github actions, which I believe is unrelated.

Electron100 (approved):

PR #164, opened 2023-11-07
Support BTreeMap for feature json

jayvdb:


butane_core/src/codegen/mod.rs
```diff
@@ -404,7 +404,7 @@ pub fn get_deferred_sql_type(ty: &syn::Type) -> DeferredSqlType {
         .or_else(|| get_autopk_sql_type(ty))
         .unwrap_or_else(|| {
             DeferredSqlType::Deferred(TypeKey::CustomType(
-                ty.clone().into_token_stream().to_string(),
+                ty.clone().into_token_stream().to_string().replace(' ', ""),
```
jayvdb, line 407: another approach here would be to trim any `foo::bar::` namespace prefixes, so that there can only be a single `Baz` in any namespace.

butane_core/src/migrations/adb.rs
```diff
@@ -120,10 +132,18 @@ impl TypeResolver {
     fn find_type(&self, key: &TypeKey) -> Option<TypeIdentifier> {
         #[cfg(feature = "json")]
         if let TypeKey::CustomType(ct) = key {
-            if ct.starts_with("HashMap < String,") {
-                return Some(TypeIdentifier::from(SqlType::Json));
+            let mut map_type_names: Vec<&str> = HASHMAP_TYNAMES.to_vec();
+            map_type_names.extend(BTREEMAP_TYNAMES.iter());
+            for string_type_name in STRING_TYNAMES {
+                for map_type_name in &map_type_names {
+                    let type_name = format!("{map_type_name}<{string_type_name},");
```
Electron100, line 139: Lines 135-139 just construct the same set of (currently) 18 strings to check each time. Always hard to make performance judgments without measuring but this *feels* wasteful. I'd be inclined to cache it in a OnceCell or something.

Electron100 (approved):

butane_core/src/migrations/adb.rs
jayvdb: ya, I'll fix this, as I want to extend this to support `indexmap` and others, so it is going to get worse. may as well fix it now.

PR #168, opened 2023-11-19
Add newtype example

jayvdb:
TODO:
- Pick a different type to use for the newtype, as json-patch is a bit obscure -- https://crates.io/crates/rust_decimal would be good.
- Add a .md walk through


Electron100 (approved):

jayvdb:
Now included in https://github.com/Electron100/butane/pull/236

PR #171, opened 2024-01-13
Improve DataObject.save() generation

jayvdb:
- Use `Self::` and `self.` more to access existing accessible values, so generated code is easier to read. i.e. less duplicate literals in the output.
  - I am assuming the compiler knows `<Self as butane::DataResult>::COLUMNS.len()` is a constant. 
- Removes an `if values.len() > 0` from the .save() for tables with an AutoPk column that has already been inserted.
  - no doubt the compiler can detect whether `values` is or is not empty, but a human needs to think about it.
- Also trim the generated code for a table with only one column which is AutoPk, and add a test case for this.

### DIFF
--- a/butane_core/src/codegen/dbobj.rs
+++ b/butane_core/src/codegen/dbobj.rs
@@ -40,11 +40,12 @@ pub fn impl_dbobject(ast_struct: &ItemStruct, config: &Config) -> TokenStream2 {
         quote!(
             if !butane::PrimaryKeyType::is_valid(self.pk()) {
                 let pk = conn.insert_returning_pk(Self::TABLE, &[], &pkcol, &[])?;
-                self.#pkident = butane::FromSql::from_sql(pk)?;
-            }
+                Some(butane::FromSql::from_sql(pk)?)
+            } else {
+                None
+            };
         )
     } else if auto_pk {
-        let pkident = pk_field.ident.clone().unwrap();
         let values_no_pk: Vec<TokenStream2> = push_values(ast_struct, |f: &Field| f != &pk_field);
         let save_cols = columns(ast_struct, |f| !is_auto(f) && f != &pk_field);
         quote!(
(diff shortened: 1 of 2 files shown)


butane_core/src/codegen/dbobj.rs
```diff
@@ -36,7 +36,14 @@ pub fn impl_dbobject(ast_struct: &ItemStruct, config: &Config) -> TokenStream2 {
     let values: Vec<TokenStream2> = push_values(ast_struct, |_| true);
     let insert_cols = columns(ast_struct, |f| !is_auto(f));
 
-    let save_core = if is_auto(&pk_field) {
+    let save_core = if auto_pk && values.len() == 1 {
```
Electron100, line 39: I don't feel strongly enough to block because I don't think it's going to create problems, but I'm not sure I'm personally convinced the changes between here and line 65 are an improvement. You did reduce nesting at line 52, but at the expensive of duplicating lines 41-43, and I'm never a fan of duplicated code.

Electron100 (approved):

butane_core/src/codegen/dbobj.rs
Electron100: Thanks for reducing the reliance on literals in the generated code though. That part of this change I definitely agree is an improvement.

jayvdb: line 42
```rust
let pk = conn.insert_returning_pk(Self::TABLE, &[], &pkcol, &[])?;
```
and line 68
```rust
let pk = conn.insert_returning_pk(Self::TABLE, &[#insert_cols], &pkcol, &values)?;
```
are different.
Hopefully now tables with only a pk (but can include many fields which are virtual columns) will allow the compiler to completely remove the generated `let mut values: Vec....` which is unused in this scenario.  I have one of these in my schema with a few Many's on it.

jayvdb: I've now de-dup-ed the `self.#pkident = ...` part also.

Is this better, or worse?

butane_core/src/codegen/dbobj.rs
```diff
@@ -36,8 +36,16 @@ pub fn impl_dbobject(ast_struct: &ItemStruct, config: &Config) -> TokenStream2 {
...
-    let save_core = if is_auto(&pk_field) {
-        let pkident = pk_field.ident.clone().unwrap();
+    let save_core = if auto_pk && values.len() == 1 {
+        quote!(
+            if !butane::PrimaryKeyType::is_valid(self.pk()) {
+                let pk = conn.insert_returning_pk(Self::TABLE, &[], &pkcol, &[])?;
+                Some(butane::FromSql::from_sql(pk)?)
+            } else {
+                None
+            };
```
jayvdb, line 46: the semi-colons are here, otherwise rustfmt mangles the indentation.

butane_core/src/codegen/dbobj.rs
Electron100: Works for me, thanks!

Electron100 (approved):

PR #174, opened 2024-01-14
Add missing migration to getting-started example

jayvdb:
This migration is the effect of https://github.com/Electron100/butane/pull/158


jayvdb:
It would be good to have https://github.com/Electron100/butane/pull/175 considered first, then I'll refresh this PR if that one is approved.  Thus avoiding revising these files multiple times.

Electron100 (approved):

Issue #62, opened 2023-04-19
Feature to use YAML in .butane


jayvdb:
Would you mind if I created a feature to use serde_yaml to use YAML instead of JSON for the files under `.butane` ?

jayvdb:
Or TOML.  Anything but JSON, really. ;-)

Electron100:
Hmm, talk me through the rationale. Why? 

I certainly won't claim that JSON is perfect, but it's human-readable and ubiquitous. TOML is better for configuration files, but that's not really what these are. I've never been a big fan of YAML due to the indentation-dependence and general complexity, but I'll grant that it's more human-writable than JSON. But these files aren't supposed to be written by humans, they're just a persisted representation of state intended to be operated on only by the butane library/cli.

Moreover, regardless of the merits of any particular format I do think there's value in having a single format. Otherwise either everybody working on a project using the non-default format has to compile their Butane CLI with special flags when installing it, or the CLI needs to support all the formats by default and try to intelligently figure out which one any given project is using. Which is certainly doable, but it seems undesirable.

None of this is to say I'm set against it, I just want to understand the why before it goes forward.

jayvdb:
The reason is JSON is terrible for being stored in a repository, especially if the file will change.  The `current` directory especially changes with each change to our schema, and I want the diff to be readable so that checking it becomes an important part of our change review process.

jayvdb:
Another example of the problem is handling merge conflicts.  Line based formats handle merge conflicts better.  This is most obvious in `.butane/migrations/current/types.json` because it doesnt use line breaks at all, but even if it does, JSON is still problematic because it cant handle "trailing commas".

Electron100:
Overall, the source control diff rationale makes sense and seems like a good motivation for a different format. I'm curious to hear more about how checking the diff of `current` would be an "an important part of [your] change review process" though. Current should be mechanically generated from the current Rust sources, so I'm not sure what a review would focus on? If you get any merge conflicts I'd generally recommend just re-compling to regenerate the contents of `current`.

jayvdb:
The code review of `current` should verify it contains exactly and only the changes which were made to the Rust structs.  Even if it will be regenerated correctly every time, the code review process will ensure everyone becomes familiar with the changes that are expected under `current`, and then becomes concerned if they dont see what is expected, and either learn more about butane if it was behaving correctly, or investigate if the unexpected change turns out to be a bug.

PR #175, opened 2024-01-14
Use pretty json in .butane files

jayvdb:
Also appends an EOL if it is missing from file contents.

Related to https://github.com/Electron100/butane/issues/62 - If we have pretty JSON, I can put off yaml/toml as pretty json is nearly suitable for storing .butane in a VCS.


butane_core/src/migrations/fsmigrations.rs
```diff
@@ -95,7 +98,14 @@ impl FsMigration {
         self.fs
             .write(&path)?
             .write_all(contents)
-            .map_err(|e| e.into())
+            .map_err(<std::io::Error as Into<Error>>::into)?;
+        if contents[contents.len() - 1] != b'\n' {
+            self.fs
```
jayvdb, line 103: This additional open is inefficient, but allows the second commit to be optional.
Still a good idea to have this to be sure the file has EOL at EOF.  Or could become an error to ensure we never forget to append the EOL before calling `write_contents`.

butane_core/src/migrations/fsmigrations.rs
```diff
@@ -68,11 +68,14 @@ impl FsMigration {
     }
 
     fn write_info(&self, info: &MigrationInfo) -> Result<()> {
-        self.write_contents("info.json", serde_json::to_string(info)?.as_bytes())
+        self.write_contents(
+            "info.json",
+            &[serde_json::to_string_pretty(info)?.as_bytes(), b"\n"].concat(),
```
Electron100, line 73: [nit] I doubt this is a terribly performance-sensitive path, and I haven't benchmarked, but my guess is that all these concats are slower than doing multiple writes on the same open file.

butane_core/src/migrations/fsmigrations.rs
Electron100: Two questions here:
1. Rather than re-opening the file, why not just assign the first writer to a local variable and do the check and second write before letting it go?
2. If you're auto-appending a newline in this function, why bother with all the concacts at the call sites?

jayvdb: As these writes are not IMO a performance-sensitive path, my intention was to only use the first commit, and discard the second.  That involved two opens, which isnt performant, but it is clear and simple.

There is no need for the second commit which uses concats, which I think add unnecessary complication.  Maybe there is a way to do the append neater in the callers, but I couldnt see an easy one.  https://github.com/serde-rs/json/issues/315 shows how it can be done.  https://github.com/serde-rs/json/issues/746 would require a lot more restructuring.

Using a variable in `write_contents()` avoids the second file open, so I am happy to go with that, and drop the second commit.

Electron100 (approved):

Issue #162, opened 2023-11-07
Strange CI failure


jayvdb:
Not sure why this occurred, but it doesnt appear related to the PR

https://github.com/Electron100/butane/actions/runs/6779692955/attempts/1?pr=161

Rerunning the workflow succeeded.

jayvdb:
Maybe there isnt enough locking around writing and reading `types.json` ?  But it is odd that it is always around byte 41/42.

This CI looks very similar https://github.com/Electron100/butane/actions/runs/7513662527/job/20455666845?pr=170 , and it was fixed on the second run.

```rust
failed to read types "/home/runner/work/butane/butane/butane/.butane/migrations/current/types.json"
error: proc-macro derive panicked
   --> butane/tests/json.rs:234:39
    |
234 | #[derive(PartialEq, Eq, Debug, Clone, FieldType, Serialize, Deserialize)]
    |                                       ^^^^^^^^^
    |
    = help: message: called `Result::unwrap()` on an `Err` value: SerdeJson(Error("trailing characters", line: 1, column: 42))
```

And I saw it locally, and then the tests passed when I ran them again. 
```rust
failed to read types "/home/jayvdb/rust/butane/butane/.butane/migrations/current/types.json"
error: custom attribute panicked
  --> butane/tests/common/blog.rs:86:1
   |
86 | #[model]
   | ^^^^^^^^
   |
   = help: message: called `Result::unwrap()` on an `Err` value: SerdeJson(Error("trailing characters", line: 1, column: 41))
```

jayvdb:
One of our team had this occur repeatedly until they deleted their `~/.cargo/registry`

jayvdb:
still occurring
https://github.com/Electron100/butane/actions/runs/8320446441/job/22765175487

jayvdb:
On Windows https://github.com/Electron100/butane/actions/runs/8494192148/job/23269030123
```
cd butane && cargo +stable test --all-features
 Downloading crates ...
  Downloaded strsim v0.7.0
  Downloaded unicode-xid v0.1.0
  Downloaded syn v0.[15](https://github.com/Electron100/butane/actions/runs/8494515035/job/23269791025?pr=232#step:16:16).44
  Downloaded quote v0.6.13
  Downloaded proc-macro2 v0.4.30
  Downloaded exec_time v0.1.4
  Downloaded darling_macro v0.9.0
  Downloaded darling_core v0.9.0
  Downloaded darling v0.9.0
[...]

PR #176, opened 2024-01-14
Fix eprintln when failing to write types.json

jayvdb:


butane_core/src/migrations/fsmigrations.rs
```diff
@@ -161,7 +161,7 @@ impl MigrationMut for FsMigration {
             TYPES_FILENAME,
             serde_json::to_string(&types)
                 .map_err(|e| {
-                    eprintln!("failed to read types");
+                    eprintln!("failed to write types");
```
jayvdb, line 164: alternatively, this line could be removed, but I think it is helpful until we switch to using `log` throughout.

e.g. the eprintln above (the real "read" error) appears at https://github.com/Electron100/butane/issues/162

Also potentially this should be
```rust
eprintln!("failed to write types {typefile:?}");
```
similar to the read error above.

Electron100 (approved):

PR #177, opened 2024-01-14
Fix sqlite foreign key support

jayvdb:
Related to https://github.com/Electron100/butane/issues/109

c.f. https://www.sqlite.org/foreignkeys.html

### DIFF
--- a/butane_core/src/db/sqlite.rs
+++ b/butane_core/src/db/sqlite.rs
@@ -58,7 +58,9 @@ impl SQLiteBackend {
 }
 impl SQLiteBackend {
     fn connect(&self, path: &str) -> Result<SQLiteConnection> {
-        SQLiteConnection::open(Path::new(path))
+        let connection = SQLiteConnection::open(Path::new(path))?;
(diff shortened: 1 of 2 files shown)


butane_core/src/db/sqlite.rs
```diff
@@ -58,7 +58,9 @@ impl SQLiteBackend {
 }
 impl SQLiteBackend {
     fn connect(&self, path: &str) -> Result<SQLiteConnection> {
-        SQLiteConnection::open(Path::new(path))
+        let connection = SQLiteConnection::open(Path::new(path))?;
+        connection.execute("PRAGMA foreign_keys = ON")?;
```
jayvdb, line 62: I wasnt sure whether this belongs here, or in the `SQLiteConnection::open()`

Electron100: I don't have a real opinion either. They're both part of the same module and this is the only caller so I'm not sure it matters.

Electron100 (approved):

PR #179, opened 2024-01-14
Fill in some missing docs

jayvdb:


butane_core/src/lib.rs
```diff
@@ -1,5 +1,9 @@
+//! Library providing functionality butane macros and tools use.
```
Electron100, line 1: I'm having trouble parsing this sentence.
Really most of the butane logic ended up in butane_core. Theoretically/originally it's for what needed to be shared with butane_codegen, but that turned out to be a lot of it, and so much of the rest fell in too.

jayvdb: "Library providing functionality used by butane macros and tools." ?

Electron100 (approved):

PR #194, opened 2024-02-28
Put reused filenames in constants.

jayvdb:


examples/getting_started/src/lib.rs
```diff
@@ -5,13 +5,15 @@
...
 
+use butane::_filenames::{BUTANE_DIRNAME, CONNECTION_JSON_FILENAME};
 use butane::db::{Connection, ConnectionSpec};
 use butane::prelude::*;
 use models::{Blog, Post};
 
 /// Load a [Connection].
 pub fn establish_connection() -> Connection {
-    butane::db::connect(&ConnectionSpec::load(".butane/connection.json").unwrap()).unwrap()
+    let connspec = format!("{BUTANE_DIRNAME}/{CONNECTION_JSON_FILENAME}");
```
jayvdb, line 15: IMO the fact that the examples need to include these strings feels like a problem.
These shouldnt need to be so visible - they should be a hidden detail that are only important for unusual usage.

butane_core/src/lib.rs
```diff
@@ -29,6 +29,15 @@ use db::{BackendRow, Column, ConnectionMethods};
 pub use query::Query;
 pub use sqlval::{AsPrimaryKey, FieldType, FromSql, PrimaryKeyType, SqlVal, SqlValRef, ToSql};
 
+/// Filenames used in butane.
+pub mod _filenames {
```
jayvdb, line 33: Using `_` to indicate semi-private nature

examples/getting_started/src/lib.rs
jayvdb: I think `ConnectionSpec::load` should auto detect if this filename exists

PR #198, opened 2024-03-06
Add CLI backend subcommand

jayvdb:
Has some minor backwards incompatibilities in `ms.create_migration`, which could be avoided, however the next release already has a few such as removal of `.state`.


butane_cli/src/lib.rs
```diff
@@ -231,6 +237,55 @@ pub fn load_connspec(base_dir: &PathBuf) -> Result<db::ConnectionSpec> {
...
+            std::process::exit(1);
+        }
+    } else {
+        // Verify named backends were in the latest migration.
+        if let Ok(ms) = get_migrations(base_dir) {
+            if let Some(latest_migration) = ms.latest() {
+                if let Ok(migration_backends) = latest_migration.sql_backends() {
+                    for backend_name in backend_names {
+                        if !migration_backends.contains(backend_name) {
+                            eprintln!("Adding new backend {backend_name} is not supported.  Use collapse to use a new backend.");
```
jayvdb, line 271: "Use collapse to use a new backend." doesnt make sense yet , as collapse doesnt allow `--backend` args.

jayvdb:
CI failure here is the same as https://github.com/diesel-rs/diesel/issues/3960

jayvdb:
CI fixed

butane_cli/src/main.rs
```diff
@@ -20,6 +21,16 @@ struct Cli {
 enum Commands {
     /// Initialize the database.
     Init(InitCommand),
+    /// Add a backend to existing migrations.
+    AddBackend {
```
jayvdb, line 25: I think I want to change this again, so it is `butane backends [add|remove|list]` reducing this functionality to a single top level sub-command

Electron100 (approved):

butane_cli/src/lib.rs
```diff
@@ -65,7 +67,9 @@ pub fn init(base_dir: &PathBuf, name: &str, connstr: &str, connect: bool) -> Res
     Ok(())
 }
 
-pub fn make_migration(base_dir: &PathBuf, name: Option<&String>) -> Result<()> {
+/// Make a migration.
+/// When `backend_names` is empty, the backend is selected from the existing migrations or connection.
```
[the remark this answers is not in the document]
jayvdb: It did exist in an earlier version - will update.

Issue #199, opened 2024-03-06
rollback fails if db is not at latest migration


jayvdb:
Steps to reproduce:

1. Create a migration "a"
2. Migrate the database to "a"
3. Create a migration "b"
4. Run rollback

rollback will attempt to rollback "b" which isnt in the database, resulting in errors.

This occurs for both `butane rollback` and `butane rollback a`

PR #200, opened 2024-03-06
Fix rollback when partially migrated

jayvdb:
Fixes https://github.com/Electron100/butane/issues/199


butane_cli/src/main.rs
```diff
@@ -55,7 +58,7 @@ However if the migration has been manually edited, it will need to be manually r
     /// Rollback migrations. With no arguments, undoes the latest migration. If the name of a migration is specified, rolls back until that migration is the latest applied migration.
     Rollback {
         /// Migration to roll back to.
-        name: String,
+        name: Option<String>,
```
jayvdb, line 61: This was a bug in https://github.com/Electron100/butane/pull/197

PR #203, opened 2024-03-08
Bump dependencies

jayvdb:


butane_core/src/db/helper.rs
```diff
@@ -239,9 +237,9 @@ pub fn column_default(col: &AColumn) -> Result<SqlVal> {
             #[cfg(feature = "json")]
             SqlType::Json => SqlVal::Json(serde_json::Value::default()),
             #[cfg(feature = "datetime")]
-            SqlType::Timestamp => {
-                SqlVal::Timestamp(NaiveDateTime::from_timestamp_opt(0, 0).unwrap())
-            }
+            SqlType::Timestamp => SqlVal::Timestamp(chrono::naive::NaiveDateTime::from(
+                chrono::NaiveDate::from_yo_opt(1970, 0).unwrap(),
```
jayvdb, line 241: I believe this matches the previous default which was UNIX epoch
https://docs.rs/chrono/latest/chrono/naive/struct.NaiveDateTime.html#method.from_timestamp

Electron100: Hmm, wouldn't the more direct conversion be `chrono::DateTime::from_timestamp(0, 0).unwrap().naive_utc()`? That way we avoid knowledge of 1970

PR #209, opened 2024-03-18
Move save logic out of procmacro

Electron100:
Generated code is challenging to write and even more challenging to debug. As part of a gradual effort to reduce the amount (or at least complexity) of generated code, move the `DataObject::save` method out of codegen, with a handful of simpler helpers needed by that method generated instead.

This is a refactoring changes. No difference is expected in API surface or result


butane_core/src/codegen/dbobj.rs
```diff
@@ -108,30 +63,41 @@ pub fn impl_dbobject(ast_struct: &ItemStruct, config: &Config) -> TokenStream2 {
     quote!(
         #dataresult
 
+        impl butane::internal::DataObjectInternal for #tyname {
+            const NON_AUTO_COLUMNS: &'static [butane::db::Column] = &[
+                                #insert_cols
```
jayvdb, line 68: dedent?

jayvdb (approved):
lgtm; one comment

PR #211, opened 2024-03-18
Fix three lints from beta & nightly

jayvdb:


butane_core/src/migrations/fsmigrations.rs
```diff
@@ -429,6 +429,7 @@ impl MigrationLock {
             .read(true)
             .write(true)
             .create(true)
+            .truncate(true)
```
jayvdb, line 432: from beta
```
warning: file opened with `create`, but `truncate` behavior not defined
   --> butane_core/src/migrations/fsmigrations.rs:431:14
    |
431 |             .create(true)
    |              ^^^^^^^^^^^^- help: add: `.truncate(true)`
    |
    = help: if you intend to overwrite an existing file entirely, call `.truncate(true)`
    = help: if you instead know that you may want to keep some parts of the old file, call `.truncate(false)`
    = help: alternatively, use `.append(true)` to append to the file instead of overwriting it
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_open_options
    = note: `#[warn(clippy::suspicious_open_options)]` on by default
```

butane_core/src/migrations/memmigrations.rs
```diff
@@ -125,7 +125,7 @@ impl Default for MemMigrations {
 impl Migrations for MemMigrations {
     type M = MemMigration;
     fn get_migration(&self, name: &str) -> Option<Self::M> {
-        self.migrations.get(name).map(MemMigration::clone)
+        self.migrations.get(name).cloned()
```
jayvdb, line 128: from beta - self-evident

butane_core/src/migrations/mod.rs
```diff
@@ -281,7 +281,7 @@ impl DataObject for ButaneMigration {
 
 impl crate::internal::DataObjectInternal for ButaneMigration {
     const NON_AUTO_COLUMNS: &'static [Column] = Self::COLUMNS;
-    fn pk_mut(&mut self) -> &mut String {
+    fn pk_mut(&mut self) -> &mut impl PrimaryKeyType {
```
jayvdb, line 284: from nightly
```
warning: impl trait in impl method signature does not match trait method signature
   --> butane_core/src/migrations/mod.rs:284:29
    |
284 |     fn pk_mut(&mut self) -> &mut String {
    |                             ^^^^^^^^^^^
    |
   ::: butane_core/src/lib.rs:70:33
    |
70  |         fn pk_mut(&mut self) -> &mut impl PrimaryKeyType;
    |                                 ------------------------ return type from trait method defined here
    |
    = note: add `#[allow(refining_impl_trait)]` if it is intended for this to be part of the public API of this crate
    = note: we are soliciting feedback, see issue #121718 <https://github.com/rust-lang/rust/issues/121718> for more information
    = note: `#[warn(refining_impl_trait_internal)]` on by default
help: replace the return type so that it matches the trait
    |
284 |     fn pk_mut(&mut self) -> &mut impl sqlval::PrimaryKeyType {
    |                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Electron100 (approved):

PR #214, opened 2024-03-18
Remove eprintln of query sql

jayvdb:
Also
- Remove eprintln used in tests


jayvdb:
Interesting - using +stable like https://github.com/Electron100/butane/pull/214/commits/5fe3704bc2caf5e547f47cebd547c060d4fc29e1causes 0 tests to be run

PR #223, opened 2024-03-24
Regenerate migrations

jayvdb:


examples/getting_started/tests/rollback.rs
```diff
@@ -0,0 +1,23 @@
...
+    // Migrate forward.
+    let base_dir = std::path::PathBuf::from(".butane");
+    let migrations = butane_cli::get_migrations(&base_dir).unwrap();
+    let to_apply = migrations.unapplied_migrations(&connection).unwrap();
+    for migration in &to_apply {
+        migration.apply(&mut connection).unwrap();
+        eprintln!("Applied {}", migration.name());
+    }
+
+    // TODO: Insert data
```
jayvdb, line 15: I expect this isnt do-able atm
https://github.com/Electron100/butane/pull/158 probably broke rollback, as the order of tables to be removed needs to be adjusted.
If it does work, it might be because the alpha order of these tables hides the problem.
Needs investigating

jayvdb: Raised https://github.com/Electron100/butane/issues/226

PR #224, opened 2024-03-24
Split clear_migrations

jayvdb:


butane_core/src/migrations/mod.rs
```diff
@@ -130,7 +130,17 @@ where
     /// any storage backing it) and deleting the record of their
```
Electron100, line 130: The "deleting the record of their existence/application from the database" line here is no longer accurate, so this comment needs an update.

Could you add an example of when  someone might choose to use `delete_migrations` instead of `clear_migrations`? I don't object to it, just not immediately seeing the use case.

Electron100 (changes requested):

butane_core/src/migrations/mod.rs
jayvdb: https://github.com/Electron100/butane/pull/223 (merged without your approval) and https://github.com/Electron100/butane/pull/225 are examples of where deleting migrations on disk is desirable without touching the database.

Electron100 (approved):

Issue #121, opened 2023-06-26
Migrations directory contains lots of duplicate .table files


jayvdb:
Each migration directory contains .table files for every model.  I suspect it only needs to contain .table files for models that were modified in that migration.  The `current` migration has the current complete set, and a complete set could also be obtained by stepping through the previous migrations back to the initial migration.

Electron100:
That's true, although there's a reasonable bit of refactoring in the migrations logic that would be necessary to do that space optimization. Are you observing a particular problem from the duplicate files, or just observing it as an area for general cleanliness improvement?

jayvdb:
It is just an observation, and not important.  Similar to https://github.com/Electron100/butane/issues/62 , when these files are being committed and reviewed, if there is lots of unnecessary .table files in a PR it is harder to carefully review the .table files which are modified.

PR #225, opened 2024-03-25
Avoid duplicate .table files

jayvdb:
Closes https://github.com/Electron100/butane/issues/121


butane_core/src/migrations/fsmigrations.rs
```diff
@@ -20,13 +20,17 @@ const TYPES_FILENAME: &str = "types.json";
 struct MigrationInfo {
     /// The migration this one is based on, or None if this is the
     /// first migration in the chain
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     from_name: Option<String>,
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    existing_schema: BTreeMap<String, String>,
```
jayvdb, line 26: im not especially happy with this name.

butane_core/src/migrations/fsmigrations.rs
```diff
@@ -135,16 +145,74 @@ impl FsMigration {
...
         )
     }
 
+    fn add_unmodified_table(
+        &mut self,
+        table: &ATable,
+        from_migration: &impl Migration,
+    ) -> Result<()> {
+        // It isnt possible to use from_migration.info() as that isn't part of `Migration` trait.
+        // So first step is to convert to FsMutation.
```
jayvdb, line 191: https://github.com/Electron100/butane/pull/225/commits/61d046b8137736da678a6f22805077838d4853c3#diff-4d962e128c62a0c6509b79195b8c55a41d00c21463a4eed3ef3764d27aea2ec0R173 is another approach which doesnt convert the arg into a `FsMutation` ; i.e. only use `Mutation`

butane_core/src/migrations/fsmigrations.rs
```diff
@@ -135,16 +145,74 @@ impl FsMigration {
...
         self.write_contents(
             &format!("{}.table", table.name),
             serde_json::to_string_pretty(table)?.as_bytes(),
         )
     }
 
+    fn add_unmodified_table(
+        &mut self,
+        table: &ATable,
+        from_migration: &impl Migration,
```
jayvdb, line 188: this could be `from_migration_name: String` as that is all that is currently used

butane_core/src/migrations/fsmigrations.rs
Electron100: Perhaps `table_bases`

butane_core/src/migrations/fsmigrations.rs
Electron100, line 191: Here and below, do you mean `FsMigration` when you say FsMutation?

butane_core/src/migrations/fsmigrations.rs
```diff
@@ -135,16 +145,74 @@ impl FsMigration {
...
+    ) -> Result<()> {
+        // It isnt possible to use from_migration.info() as that isn't part of `Migration` trait.
+        // So first step is to convert to FsMutation.
+        let migrations_dir = self.root.parent().unwrap();
+        let migrations = crate::migrations::from_root(migrations_dir);
+        let migration_list = migrations.all_migrations()?;
+        let from_mutation = migration_list
```
Electron100, line 195: [nit] I don't love using the name "mutation" for a variable which represents a migration. I think you're doing it because there's already an existing variable named `from_migration`. I'd prefer doing one of the following
1. Assign the name to another intermediate variable and then alias `from_migration`
2. Consider a name such as `from_fsmigration` which imo better conveys the difference between the two variables (one is an `FsMigration` instead of just a `Migration`

butane_core/src/migrations/fsmigrations.rs
```diff
@@ -135,16 +145,74 @@ impl FsMigration {
...
+            .iter()
+            .find(|&m| m.name() == from_migration.name())
+            .unwrap();
```
Electron100, line 198: This method returns a `Result` so prefer early-return of an error to `unwrap`. Same comment on the panic at line 201.

butane_core/src/migrations/migration.rs
```diff
@@ -86,8 +86,19 @@ pub trait MigrationMut: Migration {
     /// Adds an abstract table to the migration. The table state should
     /// represent the expected state after the migration has been
     /// applied. It is expected that all tables will be added to the
-    /// migration in this fashion.
-    fn write_table(&mut self, table: &ATable) -> Result<()>;
+    /// migration in this fashion, if they were modified in this migration.
+    fn add_modified_table(&mut self, table: &ATable) -> Result<()>;
```
Electron100, line 90: I think I would prefer not to rename this method. I'm open to the argument that the new name is better/clearer but I still prefer not to break backwards compatibility solely to clarify a method name.

butane_core/src/migrations/migration.rs
```diff
@@ -86,8 +86,19 @@ pub trait MigrationMut: Migration {
...
+
+    /// Marks a table as not modified in this migration.
+    /// Use instead of `add_modified_table`.
+    #[allow(unused_variables)]
+    fn add_unmodified_table(
+        &mut self,
+        table: &ATable,
+        from_migration: &impl Migration,
+    ) -> Result<()> {
+        self.add_modified_table(table)
```
Electron100, line 100: This is a surprising default I think. Since we don't expect a lot of custom implementations of Migration, do we really need this default?

Electron100 (changes requested):

butane_core/src/migrations/fsmigrations.rs
jayvdb: I was also thinking a name including "inherit" might work.

butane_core/src/migrations/migration.rs
jayvdb: IMO now is the time to do any breaking changes.

We've already broken it with the removal of `.state`, and so far there is only one dependent crate https://github.com/DimmKG/butane_rocket_pool per https://crates.io/crates/butane/reverse_dependencies , and one other project https://github.com/TmLev/crud-grpc-rust-tonic-butane-postgres using it per https://github.com/search?q=butane+language%3Arust&type=repositories .  Code search reveals one more : https://github.com/janosimas/code_as_db , and I think I found another one a year ago, but cant find it now.

Electron100: Public OSS doesn't necessarily tell they true number of dependents, but you're certainly right that it's not high. 
To my mind this is one of the less necessary breaking changes, but on the other hand the chances that it even breaks any dependents is very slim, since it's primarily for use of butane itself and not needed by ordinary projects.
So I won't block it going forward.

Electron100 (approved):

PR #228, opened 2024-03-27
Log directory deletions

jayvdb:


butane_core/src/migrations/fsmigrations.rs
```diff
@@ -371,6 +371,8 @@ impl MigrationsMut for FsMigrations {
         &mut self.current
     }
     fn clear_current(&mut self) -> Result<()> {
+        #[cfg(feature = "log")]
```
Electron100, line 374: I don't feel strongly about this, but the reason `lib.rs` has the `pub use` and the no-op macro is so log statements don't all have to be decorated with `#cfg`

Electron100 (approved):

butane_core/src/migrations/fsmigrations.rs
jayvdb: Thanks.  Fixed.

PR #229, opened 2024-03-29
Various Dev Experience improvements

jayvdb:


.editorconfig
```diff
@@ -0,0 +1,16 @@
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+max_line_length = 120
+
+[butane_cli/src/main.rs]
+max_line_length = 400
```
jayvdb, line 10: This file allows use of https://github.com/editorconfig-checker/editorconfig-checker .  I can add to CI if desirable.
It catches stuff that rustfmt silently ignores: https://github.com/rust-lang/rustfmt/issues/5683

Electron100: I'd be happy to have editorconfig-checker as part of CI

Electron100 (approved):

PR #230, opened 2024-03-29
cli: Describe migration

jayvdb:


butane_cli/src/lib.rs
```diff
@@ -92,6 +93,125 @@ pub fn make_migration(base_dir: &Path, name: Option<&String>) -> Result<()> {
...
+        }
+    }
+    Ok(())
+}
+
+/// Print a description of the current migration.
+pub fn describe_current_migration(base_dir: &Path) -> Result<()> {
+    let mut ms = get_migrations(base_dir)?;
+    let to_db = ms.current().db()?;
+    let mut from_db = ADB::new();
```
Electron100, line 178: [nit] mutable is unnecessary here, the `ADB::new()` can be in an else clause (at line 181)

butane_cli/src/lib.rs
```diff
@@ -92,6 +93,125 @@ pub fn make_migration(base_dir: &Path, name: Option<&String>) -> Result<()> {
...
+    if old.default() != new.default() {
+        println!("  default: {:?} -> {:?}", old.default(), new.default());
+    }
+    if old.reference() != new.reference() {
+        let old = match old.reference() {
+            Some(ARef::Literal(reference)) => {
+                format!("{}.{}", reference.table_name(), reference.column_name())
+            }
+            None => "None".to_string(),
+            Some(ARef::Deferred(_)) => panic!(),
```
Electron100, line 122: Not blocking, but I think I would prefer not panicking here. In practice `Migration::db` resolves all types, and there will probably be other things that break if it doesn't, but the method description/contract doesn't actually promise that

If you keep the panic, I'd prefer adding a panic message indicating why this should be impossible (in this case because `Migration::db` resolves types

butane_cli/src/lib.rs
```diff
@@ -92,6 +93,125 @@ pub fn make_migration(base_dir: &Path, name: Option<&String>) -> Result<()> {
...
+        println!("No changes");
+        return Ok(());
+    }
+    for op in &ops {
+        use Operation::*;
+        match op {
+            AddTable(table) | AddTableIfNotExists(table) => {
+                println!("New table {}", table.name);
+                for column in &table.columns {
+                    println!("  {} -> {:?}", column.name(), column.typeid()?);
```
Electron100, line 148: Consider implementing `Display` for `AColumn` and then just printing the column here.
In the current formatting, I don't understand what the arrow is intended to convey. It seems to have a very different meaning than it does in `print_column_diff`

jayvdb: ya, ":" would be better here.
Will implement `Display`.
Also thinking about implementing one of https://crates.io/search?q=indent to do the indentation.

jayvdb: I've changed this to `:`.

IMO `Display` isnt appropriate here.  `Display` should display all the details of the column, whereas I want to only summarise it here, so-as to not drown the user with info they already know.  If the user wants all the details, they can (and should) study the SQL.

Electron100: Ok, that's reasonable

Electron100 (approved):

PR #233, opened 2024-04-01
CI: Add editorconfig-checker

jayvdb:


jayvdb:
Raised https://github.com/editorconfig-checker/editorconfig-checker/issues/330 upstream.  If there isnt a fix soon, I'll disable the step on Windows.

PR #234, opened 2024-04-01
Add equality of KnownId(t) and Known(t)

jayvdb:
`20240115_023841384_dbconstraints` in the getting_started example had all columns being modified because the `.sqltype` members of the previous migration was using `Known(t)` while the new .table files used `KnownId(t)`.

This could be achieved by adding a different equality-like method, rather than using `.eq()`, but it seems this is a good approach for handling this deprecation.


butane_core/src/migrations/adb.rs
```diff
@@ -327,7 +327,7 @@ impl ATable {
 }
 
 /// SqlType which may not yet be known.
-#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[derive(Clone, Debug, Deserialize, Serialize)]
```
Electron100, line 330: Did you intend to remove `Eq`? If so, why?

butane_core/src/migrations/adb.rs
```diff
@@ -351,6 +351,32 @@ impl DeferredSqlType {
...
+                    return *sqltype == *other_sqltype;
+                }
+                // Replace self with KnownId, and run .eq again
+                other.eq(&Self::KnownId(sqltype.clone().into()))
+            }
+            Self::KnownId(type_id) => {
+                if let Self::KnownId(other_type_id) = other {
+                    return *type_id == *other_type_id;
+                }
+                false
```
Electron100, line 369: Don't we need to check if other is `Known`?

jayvdb: I've added a manual `Eq` impl.
I thought there was a clippy lint that ensured that if `PartialEq` is manually implemented, `Eq` should also be manually implemented.  However I cant see it now in https://rust-lang.github.io/rust-clippy/master/index.html#/derive%20partialeq :/
Happy to switch back to deriving `Eq`.

Electron100: Deriving Eq seems slightly more succinct and produces the same result, but I don't feel strongly

butane_core/src/migrations/adb.rs
```diff
@@ -351,6 +351,38 @@ impl DeferredSqlType {
         }
     }
 }
+/// Compare, with Known and KnownId being identical if they contain the same type.
+impl PartialEq<DeferredSqlType> for DeferredSqlType {
+    fn eq(&self, other: &DeferredSqlType) -> bool {
+        match self {
```
Electron100, line 357: I think there's a more succinct formulation of this match statement. See https://github.com/Electron100/butane/commit/3a1e07cafcd6d2e580105551be772afb17034660 if you'd like to adopt it

Electron100 (approved):

PR #241, opened 2024-04-11
Add more Clones

jayvdb:


examples/getting_started/tests/rollback.rs
```diff
@@ -40,16 +40,6 @@ fn migrate_and_rollback(mut connection: Connection) {
     let migrations = butane_cli::get_migrations(&base_dir).unwrap();
     let to_apply = migrations.unapplied_migrations(&connection).unwrap();
     for migration in &to_apply {
-        if connection.backend_name() == "pg"
-            && migration.name() == "20240115_023841384_dbconstraints"
-        {
-            // migration 20240115_023841384_dbconstraints failed: Postgres error db error:
-            // ERROR: cannot drop table tag because other objects depend on it
-            // DETAIL: constraint post_tags_many__butane_tmp_has_fkey1 on table post_tags_many depends on table tag
-            let err = migration.apply(&mut connection).unwrap_err();
```
jayvdb, line 49: I've no idea why adding Clones would mean that the migration now succeeds, but then the rollback fails.

Electron100: There's something seriously spooky going on here.  On my machine on a clean `master`, this test fails. Line 49 fails because the `apply` succeeds, so `unwrap_err` panics. Something is nondeterministic here that shouldn't be, but I haven't been able to figure out what yet.

jayvdb: I see the same behaviour as CI.

Electron100: I'm signing off because this change doesn't make any of the weirdness worse.  While I still don't understand why I'm seeing different behavior from the CI, or why deriving Clone changed your/CI behavior, I do think the fact that we're currently handling all column changes with a recreate-then-rename off the table makes the generated SQL much harder to read and also makes the constraints harder to handle properly. I'm working on improving that, but didn't quite finish in the 45 minutes I was able to grab tonight. I'll try to post it tomorrow.

Electron100 (approved):

PR #242, opened 2024-04-15
Fix nightly rustc warnings in generated impls

jayvdb:
I use https://github.com/dtolnay/cargo-expand quite a bit, and the generated code atm emits lots of "redundant import", "mut not needed", etc.

### DIFF
--- a/butane_core/src/codegen/dbobj.rs
+++ b/butane_core/src/codegen/dbobj.rs
@@ -56,6 +56,34 @@ pub fn impl_dbobject(ast_struct: &ItemStruct, config: &Config) -> TokenStream2 {
         })
         .collect();
 
+    let conn_arg_name = if many_save.is_empty() {
+        syn::Ident::new("_conn", Span::call_site())
+    } else {
+        syn::Ident::new("conn", Span::call_site())
+    };
+
+    let non_auto_values_fn = if values.is_empty() {
+        quote!(
+            fn non_auto_values(&self, _include_pk: bool) -> Vec<butane::SqlValRef> {
+                return vec![];
+            }
+        )
+    } else {
+        quote!(
+            fn non_auto_values(&self, include_pk: bool) -> Vec<butane::SqlValRef> {
+                let mut values: Vec<butane::SqlValRef> = Vec::with_capacity(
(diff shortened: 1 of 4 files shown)


butane_core/src/codegen/dbobj.rs
```diff
@@ -56,6 +56,34 @@ pub fn impl_dbobject(ast_struct: &ItemStruct, config: &Config) -> TokenStream2 {
...
 
+    let conn_arg_name = if many_save.is_empty() {
+        syn::Ident::new("_conn", Span::call_site())
+    } else {
+        syn::Ident::new("conn", Span::call_site())
+    };
+
+    let non_auto_values_fn = if values.is_empty() {
+        quote!(
+            fn non_auto_values(&self, _include_pk: bool) -> Vec<butane::SqlValRef> {
```
jayvdb, line 67: I suggest renaming this, as despite me reviewing the introduction of this recently, I spent way too much time wondering why it was generating "fn values()"  as `if include_pk { } else { }`

Electron100: I'm afraid `non_auto_values` seems more confusing to me, because it seems to promise a restriction that isn't there. If the primary key is `auto` and you call `non_auto_values(true)`, it will happy return an auto-valued pk. 
The principle here is that a method's name should indicate what it does, not how it's expected to be used. 

Could we do one of the following? Either
1. Just add some comments to this method and keep the `values` name
2. Instead of making `include_pk` a parameter, split into two methods, such as `all_values` and `values_except_pk`
?

jayvdb: >  If the primary key is auto and you call non_auto_values(true), it will happy return an auto-valued pk.

I believe it does not.  Have a look at `push_values`.  It is only used in saving, and autos dont need to be saved (new creates new value; update shouldnt modify pk)

That said, I do like the idea of splitting `values()` so the names can be more clear.

Electron100: Oh, you're right, sorry. So splitting might be clearer, but I don't feel strongly.

Electron100 (approved):

PR #243, opened 2024-04-16
[pg] Use ALTER TABLE instead of copies

Electron100:
For the Postgres backend, use the appropriate `ALTER TABLE` command for changed columns instead of copying and recreating the table.

This makes the generated sql more readable and also makes it easier to be correct with foreign key constraints, since the recreated table does not automatically replace the old one in constraints.


Electron100:
TODO: needs unit tests

jayvdb (approved):

PR #250, opened 2024-04-20
Add migrate & unmigrate core helpers

jayvdb:
These helpers could be placed on `Migrations` or on `BackendConnection`, but it feels sensible to keep them separate and use these helpers to join them.  Keen to reconsider that.

My reason for keeping them simple is anyone invoking these helpers can also play with the connection and migrations in advance, either to detect any problems in advance, or after, and editing the migrations to control the start / end of the migrate/rollback operations.

I'm also not keen on trying to use these helpers in `butane_cli`, as it would require moving a lot of CLI-ish feedback into an `Err`, but I am happy to do that - probably by creating a new `Error` enum for this.

### DIFF
--- a/butane_core/src/migrations/mod.rs
+++ b/butane_core/src/migrations/mod.rs
@@ -326,3 +326,37 @@ impl crate::internal::DataObjectInternal for ButaneMigration {
         Ok(()) // no-op
     }
 }
+
+/// Migrate connection forward.
+pub fn migrate<M: Migration>(
+    connection: &mut impl crate::db::BackendConnection,
+    migrations: &impl Migrations<M = M>,
+) -> Result<()> {
+    let to_apply = migrations.unapplied_migrations(connection)?;
+    for migration in &to_apply {
+        crate::info!("Applying migration {}", migration.name());
+        migration.apply(connection)?;
+    }
+    Ok(())
+}
+
+/// Rollback all applied migrations.
+pub fn rollback<M: Migration>(
+    connection: &mut impl crate::db::BackendConnection,
+    migrations: &impl Migrations<M = M>,
+) -> Result<()> {
+    let mut migration = match migrations.last_applied_migration(connection)? {
+        Some(migration) => migration,
+        None => return Ok(()),
+    };
+    migration.downgrade(connection)?;
+
+    while let Ok(Some(migration_name)) = migration.migration_from() {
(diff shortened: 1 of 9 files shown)


butane_core/src/migrations/mod.rs
```diff
@@ -326,3 +326,37 @@ impl crate::internal::DataObjectInternal for ButaneMigration {
...
+    let to_apply = migrations.unapplied_migrations(connection)?;
+    for migration in &to_apply {
+        crate::info!("Applying migration {}", migration.name());
+        migration.apply(connection)?;
+    }
+    Ok(())
+}
+
+/// Rollback all applied migrations.
+pub fn rollback<M: Migration>(
```
jayvdb, line 344: perhaps this should be called `downgrade` to be consistent and avoid conflicting terminology with Transaction.rollback .

However the CLI uses `rollback`.  perhaps that should also be changed.

That said, "downgrade" doesnt sound right to me.  It feels more like a large version thing, not a single migration thing.

And the opposite of downgrade should be upgrade, not "migrate".  Perhaps "unmigrate" is clearer?  Or "revert"? or "unapply"?

As a refresher, django uses `migrate app.old_migration_name` to remove migrations.

Electron100:
Thanks for adding this, I too had been noticing it would be useful.
I think I would vote for placing these methods on `Migrations` because of the parallelism with `apply` on `Migration`. You can `apply` a single Migration, or `migrate` a whole `Migrations` collection.
It's also slightly more ergonomic to use with

`migrations.migrate(&mut connection)`

vs 

`migrations::migrate(&mut connection, &migrations)`

I could also see naming the method `apply_all` or `apply_unapplied`, but very much don't feel strongly.

butane_core/src/migrations/mod.rs
Electron100: Good question. I think `unmigrate` seems clearest to me, or maybe `unapply_all`.

jayvdb: After this PR, I'll update the CLI to use whatever terminology we adopt here, deprecating "rollback".

jayvdb: re `apply_all` or `apply_unapplied`, I'd prefer we stick with single words, and "migrate" is very common

Electron100 (approved):

PR #253, opened 2024-04-20
Drop constraints before dropping tables

jayvdb:
Allows https://github.com/Electron100/butane/pull/252 to pass

### DIFF
--- a/butane_core/src/db/pg.rs
+++ b/butane_core/src/db/pg.rs
@@ -551,9 +551,22 @@ fn sql_for_op(current: &mut ADB, op: &Operation) -> Result<String> {
         Operation::AddTableConstraints(table) => Ok(create_table_constraints(table)),
         Operation::AddTableIfNotExists(table) => Ok(create_table(table, true)?),
         Operation::RemoveTable(name) => Ok(drop_table(name)),
+        Operation::RemoveTableConstraints(table) => remove_table_constraints(table),
         Operation::AddColumn(tbl, col) => add_column(tbl, col),
         Operation::RemoveColumn(tbl, name) => Ok(remove_column(tbl, name)),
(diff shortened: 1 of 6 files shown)


butane_core/src/db/pg.rs
```diff
@@ -583,6 +596,16 @@ fn create_table_constraints(table: &ATable) -> String {
         .join("\n")
 }
 
+fn remove_table_constraints(table: &ATable) -> Result<String> {
```
Electron100, line 599: Method name is a bit misleading. It doesn't remove all constraints, it removes only the foreign key constraints. Not unique/primary key/etc.

butane_core/src/db/pg.rs
```diff
@@ -551,9 +551,22 @@ fn sql_for_op(current: &mut ADB, op: &Operation) -> Result<String> {
         Operation::AddTableConstraints(table) => Ok(create_table_constraints(table)),
         Operation::AddTableIfNotExists(table) => Ok(create_table(table, true)?),
         Operation::RemoveTable(name) => Ok(drop_table(name)),
+        Operation::RemoveTableConstraints(table) => remove_table_constraints(table),
```
Electron100, line 554: Why does this need to be a separate operation semantically? Why not just do it automatically as part of `RemoveTable` handling? It looks like ADB `diff` is always pairing them anyway, and I would argue that whether a table's constraints have to be dropped before the table can be dropped is a backend-specific attribute. Theoretically there could be a backend for which that wasn't true.

Electron100: Would have to change the signature of `RemoveTable` slightly, but I think that's OK

jayvdb: We need to process all of the tables twice - one to drop their fkey constraints , and then the tables can be dropped.

Electron100 (approved):

PR #260, opened 2024-04-28
Adopt unmigrate for schema reverts

jayvdb:
Implementing decision at https://github.com/Electron100/butane/pull/250#discussion_r1573346999

### DIFF
--- a/butane_cli/src/lib.rs
+++ b/butane_cli/src/lib.rs
@@ -309,14 +309,14 @@ pub fn unmigrate_to(base_dir: &Path, mut conn: Connection, to: &str) -> Result<(
         });
 
     if to_migration == latest {
-        eprintln!("That is the latest migration. Not schema change required.");
+        eprintln!("That is the latest migration. No schema change required.");
         std::process::exit(1);
     }
 
     let mut to_unapply = ms.migrations_since(&to_migration)?;
     if to_unapply.is_empty() {
         return Err(anyhow::anyhow!(
-            "That is the latest migration. Not schema change required."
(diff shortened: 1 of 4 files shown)


butane_cli/src/main.rs
```diff
@@ -73,8 +73,9 @@ However if the migration has been manually edited, it will need to be manually r
     },
     /// Embed migrations in the source code.
     Embed,
-    /// Rollback migrations. With no arguments, undoes the latest migration. If the name of a migration is specified, rolls back until that migration is the latest applied migration.
-    Rollback {
+    /// Undo migrations. With no arguments, undoes the latest migration. If the name of a migration is specified, rolls back until that migration is the latest applied migration.
+    #[command(alias = "rollback")]
+    Unmigrate {
```
jayvdb, line 78: The alternative here is to adopt the django approach of unmigrate being `migrate <older_migration_name`.
IMO it is better to keep them separate, so the command name clearly communicates that this is a destructive operation.

butane_cli/src/lib.rs
```diff
@@ -309,15 +309,14 @@ pub fn rollback_to(base_dir: &Path, mut conn: Connection, to: &str) -> Result<()
         });
 
     if to_migration == latest {
-        eprintln!("That is the latest applied migration, not rolling back to anything.");
+        eprintln!("That is the latest migration. Not schema change required.");
```
Electron100, line 312: [nit] "No schema change..."

Electron100 (approved):

Issue #285, opened 2024-12-04
More painful than necessary to get started


dhbradshaw:
Here was my experience this morning experimenting with butane.

1. Read the book, which pointed to version 6.
2. Read crates, which has 6.0 showing with an error and 6.1 preferred but 7 out.
3. Added version 7 to cargo and tried following the examples in the github repo.
4. Found that the example code failed with the published version 7 and tinkered around without success to get it succeeding.
5. Switched examples to the get started without async example.
6. Looked for version branches or links where I could conveniently get working, stable examples.
7. Guessed after failing to find any that while butane looks like exactly what I want it's not yet aimed at supporting external use.

I think all I was looking for was a published version for which I could easily find working examples with stable, published dependencies.

Electron100:
Thank you for the feedback. I'm fixing your first point in #285. 

The remaining difficulty you had may stem from using the published `0.7` crate but looking at the examples from latest `master`. To see the code corresponding to the `0.7` release, you can check out to the `0.7` tag in git (https://github.com/Electron100/butane/releases/tag/0.7) 
The current head of `master` has some significant revisions for the upcoming `0.8` release which makes some breaking changes in order to add optional async support.

That said, if you're interested in sharing more about the specific errors you encountered, I'm certainly happy to see what I can improve.

jayvdb:
@dhbradshaw , it would great if you could re-try with v0.8 and report back any issues you faced.

PR #286, opened 2024-12-08
Fix crate version in getting_started.md

Electron100:
Getting started guide refers to version `0.6`, but `0.7` is live on crates.io
Pointed out in #285


jayvdb:
if you rebase, CI should be green

jayvdb (approved):

Electron100:
Getting started guide refers to version 0.6, but 0.7 is live on crates.io
Pointed out in https://github.com/Electron100/butane/issues/285

PR #308, opened 2025-03-31
Update dependencies

jayvdb:
Sits on top of https://github.com/Electron100/butane/pull/307


butane_cli/src/lib.rs
```diff
@@ -7,6 +7,8 @@
 //! Backwards compatibility of the library will not even be considered, as the
 //! only objective of the crate is to provide a stable CLI.
 
+extern crate alloc;
```
jayvdb, line 10: now needed by [nonempty](https://github.com/cloudhead/nonempty)

butane_core/tests/migration.rs
```diff
@@ -209,9 +212,9 @@ fn migration_add_field_with_default_sqlite() {
         &mut sqlite_connection(),
         "ALTER TABLE Foo ADD COLUMN baz INTEGER NOT NULL DEFAULT 42;",
         // See comments on migration_add_field_sqlite
-        r#"CREATE TABLE Foo__butane_tmp (id INTEGER NOT NULL PRIMARY KEY,bar TEXT NOT NULL);
-            INSERT INTO Foo__butane_tmp SELECT id, bar FROM Foo;
-            DROP TABLE Foo;ALTER TABLE Foo__butane_tmp RENAME TO Foo;"#,
+        "CREATE TABLE Foo__butane_tmp (\"id\" INTEGER NOT NULL PRIMARY KEY,bar TEXT NOT NULL);
+            INSERT INTO Foo__butane_tmp SELECT \"id\", bar FROM Foo;
+            DROP TABLE Foo;ALTER TABLE Foo__butane_tmp RENAME TO Foo;",
```
jayvdb, line 217: the use of `r#..#` was unnecessary, and only occurred twice in this file.

butane_core/tests/adb.rs
```diff
@@ -373,10 +373,10 @@ async fn add_table_fkey_back_reference(conn: ConnectionAsync) {
             vec![
                 "CREATE TABLE a (",
                 "fkey INTEGER NOT NULL PRIMARY KEY,",
-                "FOREIGN KEY (fkey) REFERENCES b(id)",
+                "FOREIGN KEY (fkey) REFERENCES b(\"id\")",
```
jayvdb, line 376: [sqlparser](https://crates.io/crates/sqlparser) might be using this list now
https://www.postgresql.org/docs/current/sql-keywords-appendix.html

looks like we might be able to improve our quoting by using the finer grained lists at https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/keywords.rs ; iirc they have been added since we added this feature.

butane_cli/src/lib.rs
Electron100: Do you know why? It seems a tad odd.

Electron100 (approved):

butane_cli/src/lib.rs
jayvdb: It tries to use `alloc::vec::Vec`.

unfortunately that bit is hidden in their examples
https://github.com/cloudhead/nonempty/blob/master/src/lib.rs#L22-L24

Happy to use a different approach if there is one.

jayvdb: https://ethereum.stackexchange.com/a/159752 explains why it is needed

PR #316, opened 2025-04-07
Use nanosecond granularity on sqlite timestamp cols

jayvdb:
And revise timestamp test to require nanoseconds for sqlite, and microseconds for pg

### DIFF
--- a/butane/tests/basic.rs
+++ b/butane/tests/basic.rs
@@ -373,6 +373,8 @@ async fn cant_save_unsaved_fkey(conn: ConnectionAsync) {
 #[butane_test]
 async fn basic_time(conn: ConnectionAsync) {
     let mut now = Utc::now();
+    // Ensure we have a non-zero subsecond value so that the assert below confirms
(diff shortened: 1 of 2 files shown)


butane/tests/basic.rs
```diff
@@ -382,9 +386,14 @@ async fn basic_time(conn: ConnectionAsync) {
...
 
     let time2 = TimeHolder::get(&conn, 1).await.unwrap();
-    // Note, we don't just compare the objects directly because we
-    // lose some precision when we go to the database.
-    assert_eq!(time.utc.timestamp(), time2.utc.timestamp());
+    if conn.backend_name() == "sqlite" {
+        assert_eq!(time.utc, time2.utc);
+    } else {
+        assert_eq!(
+            time.utc.timestamp_subsec_micros(),
```
Electron100, line 393: Shouldn't this be `time.utc.timestamp_micros()`? `timestamp_subsec_micros` is only the fractional part after the second, so we aren't actually validating that the date/hour/minute/second match.

PR #320, opened 2025-04-08
Enable SQLite strict mode

jayvdb:


butane/tests/custom_pg.rs
```diff
@@ -1,5 +1,6 @@
-// We wrap everything in an inner module just so it's easier to have the feature gate in one place
-#[cfg(feature = "pg")]
+// We wrap everything in an inner module just so it's easier to have the feature gate in one place.
+// butane sqlite fails when using sqlite STRICT mode tripping up the custom type POINT.
+#[cfg(all(feature = "pg", not(feature = "sqlite")))]
```
jayvdb, line 3: so this problem means custom PG types and sqlite cant be in the same project.  Which is reasonable - if the project has types which depend on custom PG types, very likely sqlite support for that project is buggy.

Suggests that we need to further investigate how SQLite support is currently working with this custom "POINT" - on main, sqlite is at least able to parse "POINT" and not break.  Maybe sqlite behaves well, or can be made to co-exist with PG custom types.

Electron100: I'll respond with more detailed thoughts about the general problem on the issue, but for purposes of this PR, the introduction of the `not(feature = "sqlite"))` means we probably need to adjust the CI to run this test with only "pg". Currently we run all tests with --all-features.

> Suggests that we need to further investigate how SQLite support is currently working with this custom "POINT" - on main, sqlite is at least able to parse "POINT" and not break. Maybe sqlite behaves well, or can be made to co-exist with PG custom types.

The test itself is marked to only run for PG, so the only thing that runs for SQLite is the migration for the whole database, so essentially the CREATE TABLE statements. SQLite docs at https://sqlite.org/flextypegood.html say 

"When datatype names are provided, they can be just about any text. SQLite attempts to deduce the preferred datatype for the column based on the datatype name in the column definition, but that preferred datatype is advisory, not mandatory. The preferred datatype is known as the "column affinity". "

Which is how we got away with it so far. If we had a third backend that wasn't so flexible, we wouldn't have

Electron100: > we probably need to adjust the CI to run this test with only "pg"

Alternatively we break it out into its own crate. I did that in #324, although if you'd prefer I merge that afterwards I'm happy to wait.

jayvdb: That is a lovely approach.

Electron100 (approved):

Issue #150, opened 2023-10-22
Cfg Attributes on Model Fields Not Properly Handled


Electron100:
#148 duplicated the `Post` and `PostMetadata` model/dataresult types in the blog test, one for if `datatime` is enabled and one for if it is not. That shouldn't have been necessary  -- it should be possible simply to use `#cfg` attributes on the field. Unfortunately, the model attribute is processed before the cfg attribute and Butane doesn't propagate the cfg properly  and ends up creating references to the field that is ultimately omitted. We should check for the cfg attribute and if so propagate it to all references we create to the field.

jayvdb:
This can be triggered by using
```
% cargo test -p butane --no-default-features --features sqlite,async,async-adapter
...
error[E0560]: struct `Post` has no field named `pub_time`
  --> butane/tests/common/blog.rs:38:9
   |
38 |     pub pub_time: std::option::Option<NaiveDateTime>,
   |         ^^^^^^^^ `Post` does not have this field
   |
   = note: all struct fields are already assigned

error[E0609]: no field `pub_time` on type `&Post`
  --> butane/tests/common/blog.rs:38:9
   |
38 |     pub pub_time: std::option::Option<NaiveDateTime>,
   |         ^^^^^^^^ unknown field
   |
   = note: available fields are: `id`, `title`, `body`, `published`, `likes` ... and 2 others

error[E0560]: struct `PostMetadata` has no field named `pub_time`
  --> butane/tests/common/blog.rs:66:9
   |
66 |     pub pub_time: Option<NaiveDateTime>,
   |         ^^^^^^^^ `PostMetadata` does not have this field
   |
   = note: all struct fields are already assigned

For more information about this error, try `rustc --explain E0412`
```

jayvdb:
cargo-expand shows the following in the macro output
```rust
impl butane::DataResult for Post {
            type DBO = Post;
            const COLUMNS: &'static [butane::db::Column] = &[
                ...
                butane::db::Column::new(
                    "pub_time",
                    <std::option::Option<NaiveDateTime> as butane::FieldType>::SQLTYPE,
                ),
                ...
            ];
            fn from_row(row: &dyn butane::db::BackendRow) -> butane::Result<Self> {
                ...
                    pub_time: butane::FromSql::from_sql_ref(
                        row
                            .get(
                                4usize,
                                <std::option::Option<
                                    NaiveDateTime,
                                > as butane::FieldType>::SQLTYPE,
                            )?,
                    )?,
[...]

Issue #315, opened 2025-04-06
Rust migrations


jayvdb:
I have a uuid v4 primary key that I want to convert to be uuid v7 (https://github.com/Electron100/butane/issues/133), and data that needs to be migrated in both sqlite and pg databases.

While it might be possible to do that using pg extensions in pg SQL migrations, that is going to be hard to do in sqlite.  This and many other types of migrations are much easier to do in Rust code.

It should be too hard to include some Rust files from the migration folder, and run those as opposed to the generated SQL.

Issue #313, opened 2025-04-02
Dynamically maintained migrations


jayvdb:
butane CLI has a regenerate command.   The non-manually edited migrations should be able to be re-generated any time the migration generator is improved.

However it doesnt know which migrations were manually edited, and tries to replace them, which then breaks/degrades them.

I think it would be useful to add the ability for butane to do a migration without having stored SQL, at least for the "mem" migrations.  i.e. generate the SQL on the fly.  This may mean more metadata is needed, but I don't think so.

Then those migrations will automatically be improved as butane improves.

This should be opt-in, as it also means that there is a small chance that the old "dynamic" migrations might break with a future version of butane, either due to a bug or (less likely) a type of migration intentionally no longer being supported.

Alternatively, or possibly in addition, it would be good to add a comment in the SQL that it is automatically generated, so that the "regenerate" command knows that it can regenerate it, and knows to skip the hand-rolled migrations.  If noted in the SQL, IMO that comment should be stripped before it is embedded in the "mem" migrations, as it serves no purpose there.

PR #322, opened 2025-04-08
Feature gated struct fields

jayvdb:
Closes https://github.com/Electron100/butane/issues/150

IGNORE PR ATM - it can only be used to show the problem by running
```
cargo test -p butane --no-default-features --features sqlite,async,async-adapter
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,7 +107,9 @@ jobs:
       run: |
         cd butane
         # This activates the "no datetime" feature test across blog.rs and query.rs
-        cargo +stable test -p butane --test query --no-default-features --features sqlite,async,async-adapter
+        cargo +stable test \
+                      -p butane --test query \
+                      --no-default-features --features sqlite,sqlite-bundled,async,async-adapter
         # This runs all tests with all features on all backends
         cargo +stable test --all-features
     - name: Check example migrations have been updated


butane_core/src/codegen/migration.rs
```diff
@@ -46,6 +46,10 @@ fn create_atables(ast_struct: &ItemStruct, config: &dbobj::Config) -> Vec<ATable
             .clone()
             .expect("db object fields must be named")
             .to_string();
+        if name == *"pub_time" {
+            // Temporary solution to skip pub_time field in butane tests
+            continue;
+        }
```
jayvdb, line 52: This is causing the CI failure when running with the `datetime` feature enabled.

As this macro is being run with the cfg's not evaluated and thus fields removed, the .table files contain all of the fields, causing a mismatch with the ORM objects.

I think we need to extract the cfgs, and evaluate them explicitly.

jayvdb: It looks like there isnt any built in Rust support for this. c.f. https://github.com/rust-lang/rust/issues/90765

Maybe a crate which can help...

jayvdb: https://github.com/EmbarkStudios/cfg-expr looks like it can help.

jayvdb: `CARGO_CFG_FEATURE` is only available in `build.rs` , so this approach would require that there is a build.rs, and it saves `CARGO_CFG_FEATURE` somewhere so that the migration code can know which features will be active during compilation.

jayvdb: An alternative is to add a "cfg attributes" member to `AColumn` , so that the .table files describe the fact that some columns are conditional based on features.

This sounds more sensible, as then the .table file doesnt change based on which features are enabled, but the migration .sql still needs to be different depending on the features enabled.  That problem can be avoided if the migrations are generated at runtime (https://github.com/Electron100/butane/issues/313) or the migrations are written in Rust which can also use feature gates (https://github.com/Electron100/butane/issues/315).

The other approach to .sql files is that separate copies would be needed for each combination of features that are referred to in the schema definition.

PR #324, opened 2025-04-12
Refactor the custom_pg test into an example

Electron100:
Custom Pg types can interfere with applying migration to other database backends (see #320), meaning the `Point` type shouldn't be built at all when sqlite is enabled. To make this easier to manage, move it out into an example rather than a test under the main `butane` library.


examples/newtype/README.md
```diff
@@ -1,8 +1,11 @@
...
+
+Demonstrates using `derive(FieldType)` for Butane model support on the newtype pattern.
 
 To use this example, build the entire project using `cargo build` in the project root,
 and then run these commands in this directory:
 
 1. Initialise a Sqlite database using `cargo run -p butane_cli init sqlite db.sqlite`
 2. Migrate the new sqlite database using `cargo run -p butane_cli migrate`
 3. Run the commands, such as `cargo run --bin write_post_uuid`
+b
```
jayvdb, line 11: looks like this was accidentally dropped in here.

jayvdb (approved):

examples/newtype/README.md
Electron100: Whoops, thanks for catching

PR #327, opened 2025-04-14
Allow using butane_test_helper with single backend enabled

jayvdb:


jayvdb:
With this, and https://github.com/jayvdb/butane/commit/02a1ac90dbb81940fb142f1ff99e432054b0de30 , https://github.com/Electron100/butane/pull/324 will be green.

Issue #169, opened 2023-11-20
rowid


jayvdb:
sqlite: https://www.sqlite.org/rowidtable.html "rowid"
pg: https://www.postgresql.org/docs/current/ddl-system-columns.html "ctid"
oracle: https://stackoverflow.com/questions/2701782/rowid-oracle-any-use-for-it "rowid"

If we dont have any use for rowid, we could amend the sqlite DDL to use "without rowid", supported since [3.8.2](https://www.sqlite.org/releaselog/3_8_2.html) (2013-12-06) for the main tables.  We can also do this on the Many tables if we add a better constraint on them.

How could we use it?
It could be used as a "order by" column for stable ordering of query results.
It could be used internally to perform DML on rows, as rowid is quicker than primary keys.

It shouldnt be stored in the database, so it cant be used instead of a primary key, only in addition to primary keys.

jayvdb:
mysql doesnt have one? https://stackoverflow.com/questions/2258052/mysql-is-there-something-like-an-internal-record-identifier-for-every-record-in

PR #329, opened 2025-04-20
Add Backend.row_id_column

jayvdb:
Closes https://github.com/Electron100/butane/issues/169

### DIFF
--- a/butane_core/src/db/mod.rs
+++ b/butane_core/src/db/mod.rs
@@ -506,7 +506,13 @@ pub trait Backend: Send + Sync + DynClone {
     /// Butane name for the backend.
     fn name(&self) -> &'static str;
     /// Backend-dependent field name for the database's internal identifier for each row.
-    fn internal_row_insertion_id_field(&self) -> &'static str;
+    ///
+    /// It may be used to sort the rows in the order they were inserted,
+    /// however consult the backend documentation for the exact behavior.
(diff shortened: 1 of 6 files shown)


butane_core/src/db/mod.rs
```diff
@@ -503,10 +503,14 @@ impl<'bt> ConnectionMethods for Box<dyn BackendTransaction<'bt> + 'bt> {
 /// Database backend. A boxed implementation can be returned by name via [get_backend][crate::db::get_backend].
 #[async_trait]
 pub trait Backend: Send + Sync + DynClone {
+    /// Butane name for the backend.
     fn name(&self) -> &'static str;
+    /// Backend-dependent field name for the database's internal identifier for each row.
+    fn internal_row_insertion_id_field(&self) -> &'static str;
```
jayvdb, line 509: needs to return Option as same backends dont have one

jayvdb: and re "field", butane seems to prefer "column" in the naming.

Electron100: [nit] I would avoid the "internal" here, both because it's a verbose name and because it might wrongly imply it shouldn't be called by external consumers / is an unstable implementation detail. What about just `row_id_column`?

Electron100 (approved):

PR #332, opened 2025-04-30
fix(butane-cli): enable init with no .butane dir

### LINKED ISSUE #331: Unable to initialize without manually creating directories
I have [a project](https://github.com/kojiromike/caltry) I'm trying to use butane with. I'm just setting it up, but it has valid models and essentially runs. The thing is, if I delete .butane and the sqlite database entirely, and do `butane init sqlite caltry.sqlite`, I get this output:

```
$ rm -rf .butane caltry.sqlite
$ butane init sqlite caltry.sqlite
No butane migrations directory found. Add at least one model to your project and build.
```

I think that is unexpected behavior because the getting-started guide has `butane init …` as one of its first instructions, and it doesn't tell you to create any directories.

-----

Additional context: This is possibly related, butane doesn't want to recognize my models anymore. That is, if at this point I do

```
$ mkdir -p .butane/migrations
$ butane init sqlite caltry.sqlite
$ tree .butane
```

There are no errors, and I see this:

```
.butane
├── connection.json
└── migrations

2 directories, 1 file
```

Then if I run this:

```
$ butane make-migration init
No changes to migrate
$ tree .butane
.butane
├── connection.json
└── migrations
    └── current
        └── lock

3 directories, 2 files
```

Since there are a lot of models in src/models.rs, I expected to see a lot of generated migrations, but did not. I am not sure if this is because I'm doing something wrong with the model definitions, but it might be related to why the initialization is not seeing anything to do.

kojiromike:
fixes an issue where init aborts when there is no .butane/migrations dir, even though init is supposed to be the thing that creates that.

fixes #331

### DIFF
--- a/butane_cli/src/main.rs
+++ b/butane_cli/src/main.rs
@@ -147,19 +147,21 @@ fn main() {
         base_dir.push(".butane");
     }
 
-    // List any detached migrations.
-    if let Ok(ms) = get_migrations(&base_dir) {
-        if let Ok(detached_migrations) = ms.detached_migration_paths() {
-            if !detached_migrations.is_empty() {
-                eprintln!(
-                    "Ignoring detached migrations. Please delete or manually re-attach these:"
-                );
-                for migration in detached_migrations {
-                    eprintln!("- {migration}");
+    // List any detached migrations only for non-init commands
+    if !matches!(cli.command, Commands::Init(_)) {
+        if let Ok(ms) = get_migrations(&base_dir) {
+            if let Ok(detached_migrations) = ms.detached_migration_paths() {
+                if !detached_migrations.is_empty() {
+                    eprintln!(
+                        "Ignoring detached migrations. Please delete or manually re-attach these:"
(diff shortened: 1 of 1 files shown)


Electron100:
First of all, thank you for catching a rather embarrassing error that highlights a large test gap and even more so for proposing a fix. The one thing that gives me slight pause about this change is that we're assuming `init` is the only command that should be able to run with no `.butane` directory. I think that's true, but I could also envision it not remaining true forever, and the behavior of `get_migratoins` here is a bit of a footgun. I think it would be better to eliminate that footgun rather than working around it.
I've taken a stab at that in #335. Thoughts?

kojiromike:
> I think it would be better to eliminate that footgun rather than working around it. I've taken a stab at that in #335. Thoughts?

oh, 100%. I even thought about taking that approach, but my time and my rust weren't up to it at the moment. Thanks for the better solution, and also for filling a gap in rust's ecosystem with this project.

Issue #334, opened 2025-05-02
User table not getting generated in PostgreSQL


Nekidev:
Postgres already has a `user` table by default, so creating a User model in postgres doesn't actually do anything (it fails, but butane doesn't seem to notice). I could verify that the rest of my tables were properly migrated and that the user table fails to be created.

```
postgres=# CREATE TABLE "User" (
postgres(#     "id" TEXT NOT NULL PRIMARY KEY,
postgres(#     ...
postgres(# );
ERROR:  relation "User" already exists
postgres=# SELECT * FROM user;
   user   
----------
 postgres
(1 row)
```

Above, I just copy-pasted what was generated by butane and run it using `psql`. Naming tables `public."TableName"` fixes the issue.

Nekidev:
The table does get generated properly using `#[table = "user"]`.

jayvdb:
Thanks for reporting this.

We currently use https://crates.io/crates/sqlparser to know about SQL keywords which we should be quoting, and unfortunately I dont see anything in https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/dialect/postgresql.rs or https://github.com/apache/datafusion-sqlparser-rs/blob/main/src/keywords.rs which could be used to determine that "user" is an impossible table name for postgres.

IMO we would want to prevent it being used for all db backends.  However, if we were to add code to prevent it, and prevent it from sqlite also, this could break existing butane users who have been able to create a sqlite app with the table name "user".  I have confirmed that sqlite table "User" works fine.

I have created an example of this locally, and `cargo build` works - no error reported.  And it looks like I can confirm that pg migration also works, even though it is potentially silently failing.  I'll dig into this a bit more to see what is happening.

jayvdb:
https://github.com/Electron100/butane/pull/338 shows this problem

jayvdb:
#338 now has at least one fix related to this, allowing the "User" model to be created and queried.

I suspect that another fix is required.   My guess is that this

> ERROR:  relation "User" already exists

was caused by either a foreign key or a many relationship on the "User" table.

jayvdb:
I havent been able to reproduce

> ERROR: relation "User" already exists

I think I need more info about other models which could have contributed to this issue.

jayvdb:
Note I have tried to find a similar problem in SQLite.

[...]

PR #338, opened 2025-05-03
Fix quoting of table names in queries

jayvdb:
Fixes https://github.com/Electron100/butane/issues/334


Electron100:
Sorry for the slow action on this one. I'm traveling for business this week
and don't have my personal accounts, but I sign off on the change.
One non-blocking suggestion: `unmigrate.rs` doesn't look like the right
name for the test file as I don't see any unmigration happening in the
tests it contains

jayvdb:
The test name `unmigrate.rs` appears in a few of our examples, and they all attempt to run `migrations.unmigrate_async(..)`  at the end of the test, after inserting data, and it is this step which most often fails.  Definitely could be renamed to a more appropriate name.  If we do, it should be renamed for them all at the same time.

Issue #295, opened 2025-01-13
sqlite native drop column support


jayvdb:
https://www.sqlite.org/lang_altertable.html says sqlite supports `ALTER TABLE x DROP COLUMN y`.

This was added in 2021-03-12 (3.35.0) https://sqlite.org/changes.html#version_3_35_0

The bundled version of sqlite was updated to v3.35 in rusqlite two days later https://github.com/rusqlite/rusqlite/pull/917

The bundled version of sqlite is v3.46.0 as of rusqlite 0.32.0
https://github.com/rusqlite/rusqlite#notes-on-building-rusqlite-and-libsqlite3-sys

We can obtain the version of sqlite using `fn`s in https://github.com/rusqlite/rusqlite/blob/master/src/version.rs , and IMO butane could simply refuse to run if the sqlite version used is too low.

Issue #296, opened 2025-01-13
sqlite migration tmp table workaround fails due to foreign key constraints


jayvdb:
The sqlite migration generation creates tmp tables to perform table definition changes that are not supported by sqlite.

e.g. it currently uses this approach to drop columns. (c.f. https://github.com/Electron100/butane/issues/295 for a way around that)

If the table being modified is referred to by an "ForeignKey" or "Many" relationship on another table, the generated migration will fail.

e.g.

- TableA has a pk and a column foo
- TableB has a fkey to TableA
- Generate a sqlite migration to delete TableA.foo

The generated migration will do
1. create table TableA__butane_tmp ( all fields except for "foo" )
2. copy TableA data to TableA__butane_tmp for all columns except "foo"
3. drop table TableA
4. Rename table TableA__butane_tmp to TableA

At step 3, it fails with:

> DROP TABLE TableA;]: FOREIGN KEY constraint failed

jayvdb:
i.e. https://github.com/Electron100/butane/blob/31f5fbd/butane_core/src/db/sqlite.rs#L799 
fn `change_column` does not check whether there are incoming references.

Adding `PRAGMA foreign_keys = OFF` to the generated .sql doesnt appear to fix it.  That may be because sqlite docs say "This pragma is a no-op within a transaction" . A bit more detail about that at https://sqlite.org/pragma.html#pragma_foreign_keys

Electron100:
I like the use of more modern SQLite abilities. I do wonder though if when applying sqlite migrations we should just set `PRAGMA foreign_keys = OFF` for the duration. It should typically be only at program startup before any actual work happens. And there are other types of column changes that we'll potentially still need the delete-and-recreate pattern.

PR #340, opened 2025-05-04
Use SQLite 3.35 DROP COLUMN

jayvdb:
Closes https://github.com/Electron100/butane/issues/295


jayvdb:
Need to also consider https://github.com/Electron100/butane/issues/296 , possibly in this or a followup PR.

Electron100 (approved):

Issue #343, opened 2025-05-06
Allow ConnectionSpec to be created from a database URI


jayvdb:
postgresql supports connecting to a URI like `postgres://{username}:{password}@localhost:{port}/{specified_database_name}`

c.f. https://github.com/Electron100/butane/issues/306 for part of the motivation for supporting this.

If sqlite doesnt, we can manually support the typical "sqlite://" prefix, which is quite commonly used.
https://github.com/search?q=%22sqlite%3A%2F%2F%22&type=code

IMO this should be a `TryFrom` to create a `ConnectionSpec` from a `String`, and error whenever the `String` is not a URI, but it doesnt validate the database backend name is supported by butane - there are other parts of the code that provide appropriate error messages if the database backend name is not supported by butane.

And then at least the CLI would support using URI as an alternative of a backend-name&address pair.

jayvdb:
Only the CLI component is still needing implementation.

Nekidev:
`butane init pg postgres://...` worked for me. Is that what you refer to by "CLI component"?

jayvdb:
Oh good.  yes, however I was thinking that the `pg` / `sqlite` argument could now be deprecated.

Also `BUTANE_PG_CONNSTR` should be fixed to support URIs; should be a simple enough fix to butane_test_helper, where it appends the `dbname` to the connection string - this needs to be different when `BUTANE_PG_CONNSTR` is a URI

jayvdb:
`BUTANE_PG_CONNSTR` fixed to support URIs in #355

PR #344, opened 2025-05-06
Create a ConnectionSpec from a database URI

jayvdb:
First part of https://github.com/Electron100/butane/issues/343

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,6 @@ crossbeam-channel = "0.5"
 deadpool = "0.12"
 env_logger = "0.11"
 fake = "4.2"
-fluent-uri = "0.3.2"
 log = "0.4"
 maybe-async-cfg = { version = "0.2.5", default-features = false }
 nonempty = "0.11"
@@ -59,6 +58,7 @@ thiserror = "2.0"
 tokio = { version = "1"}
 tokio-postgres = "0.7"
 tokio-test = { version = "0.4"}
+url = "2.5"
 uuid = "1.2"
 
 [workspace.metadata.release]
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -38,6 +38,7 @@ crossbeam-channel = "0.5"
 deadpool = "0.12"
 env_logger = "0.11"
 fake = "4.2"
+fluent-uri = "0.3.2"
```
jayvdb, line 41: possibly should be using https://crates.io/crates/url instead.

jayvdb: `url` implements "WHATWG URL Standard", which says on https://url.spec.whatwg.org/ 
> Align RFC 3986 and RFC 3987 with contemporary implementations and obsolete the RFCs in the process. (E.g., spaces, other "illegal" code points, query encoding, equality, canonicalization, are all concepts not entirely shared, or defined.) URL parsing needs to become as solid as HTML parsing. [[RFC3986]](https://url.spec.whatwg.org/#biblio-rfc3986) [[RFC3987]](https://url.spec.whatwg.org/#biblio-rfc3987)

[`fluent-uri`](https://crates.io/crates/fluent-uri) on the other hands says
> A generic URI/IRI handling library compliant with [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) and [RFC 3987](https://datatracker.ietf.org/doc/html/rfc3987).

fwiw, I used `url` to re-implement similar functionality at work, and it was much easier to work with.  It might might be as performant as `fluent-uri`.  Not sure if that matters.  Raised https://github.com/yescallop/fluent-uri-rs/issues/35 about the perf aspect.

Electron100 (approved):

Cargo.toml
jayvdb: done

PR #348, opened 2025-05-12
Add support for dates

### LINKED ISSUE #87: Date support
`SqlValRef` currently supports `Timestamp(NaiveDateTime)` , but has no support for a date without a time component.
i.e. support for https://docs.rs/chrono/latest/chrono/naive/struct.NaiveDate.html would be helpful to some users.

sqlite has no special data type for timestamp or date.

postgres has a `date` datatype which is half the size of a `timestamp`, and also has a much later end date. c.f. https://www.postgresql.org/docs/current/datatype-datetime.html

Nekidev:
Closes #87

Adds support for dates, using PostgreSQL's `DATE` type and stored as `TEXT` in SQLite using the same format used in PostgreSQL. I didn't add support for `chrono::Date` since it's deprecated, so only `NaiveDate` is supported. Support for this is gated behind the `datetime` feature.

Usage:
```rs
use chrono::naive::NaiveDate;

#[model]
struct Table {
    pub date: NaiveDate,
}
```

TODO:
- [x] Add tests on the `TimeHolder` table at `butane/tests/basic.rs`.
- [x] I haven't been able to test this with PostgreSQL. I saw there's an `BUTANE_PG_CONNSTR` env var but when running `cargo test` with it I kept getting nonexistent test table errors, is there anything I'm missing?

### DIFF
--- a/butane/tests/basic.rs
+++ b/butane/tests/basic.rs
@@ -423,7 +423,7 @@ async fn date_as_pk(conn: ConnectionAsync) {
         .await
         .unwrap();
 
-    assert_eq!(holder.id, holder2.id);
+    assert_eq!(holder, holder2);
 }
 
 #[butane_test]
--- a/butane/tests/basic.rs
+++ b/butane/tests/basic.rs
@@ -405,7 +405,7 @@ async fn basic_time(conn: ConnectionAsync) {
 #[cfg(feature = "datetime")]
 #[model]
 #[derive(Debug, Default, PartialEq, Clone)]
-struct TimeHolderWithDatePk {
+struct DateHolder {
     pub id: NaiveDate,
 }
 
@@ -414,12 +414,12 @@ struct TimeHolderWithDatePk {
 async fn date_as_pk(conn: ConnectionAsync) {
     let now = Utc::now();
 
-    let mut holder = TimeHolderWithDatePk {
+    let mut holder = DateHolder {
         id: now.date_naive(),
     };
     holder.save(&conn).await.unwrap();
 
-    let holder2 = TimeHolderWithDatePk::get(&conn, now.date_naive())
+    let holder2 = DateHolder::get(&conn, now.date_naive())
         .await
         .unwrap();
 
--- a/butane/tests/basic.rs
+++ b/butane/tests/basic.rs
@@ -419,9 +419,7 @@ async fn date_as_pk(conn: ConnectionAsync) {
     };
     holder.save(&conn).await.unwrap();
 
-    let holder2 = DateHolder::get(&conn, now.date_naive())
-        .await
-        .unwrap();
+    let holder2 = DateHolder::get(&conn, now.date_naive()).await.unwrap();
 
     assert_eq!(holder, holder2);
 }
--- a/butane/tests/basic.rs
+++ b/butane/tests/basic.rs
@@ -6,7 +6,11 @@ use butane::{butane_type, find, find_async, model, query, AutoPk, ForeignKey};
 use butane_test_helper::*;
 use butane_test_macros::butane_test;
 #[cfg(feature = "datetime")]
-use chrono::{naive::{NaiveDate, NaiveDateTime}, offset::Utc, DateTime};
+use chrono::{
+    naive::{NaiveDate, NaiveDateTime},
+    offset::Utc,
+    DateTime,
+};
 #[cfg(feature = "sqlite")]
 use rusqlite;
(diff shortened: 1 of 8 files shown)


butane_core/src/sqlval.rs
```diff
@@ -188,6 +196,8 @@ impl fmt::Display for SqlVal {
             #[cfg(feature = "json")]
             Json(val) => f.write_str(val.as_str().unwrap()),
             #[cfg(feature = "datetime")]
+            Date(val) => val.format("%Y-%m-%d").fmt(f),
```
jayvdb, line 199: we could make `%Y-%m-%d` a constant , as it is re-used in a number of places, but IMO that isnt necessary.

butane_core/src/sqlval.rs
```diff
@@ -625,6 +639,28 @@ impl FieldType for DateTime<chrono::offset::Utc> {
...
+        SqlVal::Date(self)
+    }
+}
+#[cfg(feature = "datetime")]
+impl FieldType for NaiveDate {
+    const SQLTYPE: SqlType = SqlType::Date;
+    type RefType = str;
+}
+#[cfg(feature = "datetime")]
+impl PrimaryKeyType for NaiveDate {}
```
jayvdb, line 662: would be good to add a test in butane/tests/basic.rs for using a date as a primary key field.

jayvdb (approved):

butane_core/src/db/helper.rs
```diff
@@ -306,6 +308,8 @@ pub fn sql_literal_value(val: &SqlVal) -> Result<String> {
         #[cfg(feature = "json")]
         Json(val) => Ok(format!("{val}")),
         #[cfg(feature = "datetime")]
+        Date(val) => Ok(format!("'{}'", val.format("'%Y-%m-%d'").to_string())),
```
jayvdb, line 311: CI is complaining about the `.to_string()` here

butane_core/src/sqlval.rs
Nekidev: There is a constant which I added later and forgot to use it here, I'll update it.

Nekidev: I followed the timestamp support implementation style for this, and it used literals everywhere I saw so I did the same. I also didn't see any obvious place to place a constant, the usage of the literal isn't too big and distributed across files. If you have any suggestions about where to place such constant, I'll add it.

Electron100: I think it's fine as-is. I'd actually be a little leery of a constant because they're different things semantically.
1. In db/helper.rs we're defining the format used for a SQL literal across all SQL database types
2. In db/sqlite.rs we're defining the format used when serializing to SQLite specifically
3. Here we're defining the display format for generic printing, which may or may not be SQL.

3 certainly doesn't have to be the same as the others. 1 &2 are arguably the same. Perhaps the best thing would be for sqlite.rs to call `sql_literal_value` for types like `Date`, but I don't feel strongly.

Electron100:
Thank you for implementing this!

> I haven't been able to test this with PostgreSQL. I saw there's an BUTANE_PG_CONNSTR env var but when running cargo test with it I kept getting nonexistent test table errors, is there anything I'm missing?

You shouldn't need to set any environment variables, though you do need to have postgresql installed. What OS are you on? If you can attach a full transcript showing the command you run and the errors?

Nekidev:
I'm on windows, I was trying to run postgres using a docker container.

Nekidev:
I was able to test it with docker using the connection string formatted as KV instead of as a URL. I added a comment (out of scope, but it's helpful context). Tests passed, so that should be it.

butane/tests/basic.rs
```diff
@@ -392,6 +398,32 @@ async fn basic_time(conn: ConnectionAsync) {
...
+    let now = Utc::now();
+
+    let mut holder = DateHolder {
+        id: now.date_naive(),
+    };
+    holder.save(&conn).await.unwrap();
+
+    let holder2 = DateHolder::get(&conn, now.date_naive())
+        .await
+        .unwrap();
```
jayvdb, line 424: cargo fmt is asking for this to be
```diff
-    let holder2 = DateHolder::get(&conn, now.date_naive())
-        .await
-        .unwrap();
+    let holder2 = DateHolder::get(&conn, now.date_naive()).await.unwrap();
```

jayvdb (approved):

PR #352, opened 2025-05-16
Fix cleaning .butane when running tests

jayvdb:
`#[cfg(test)]` and `CARGO_CFG_TEST` do not work in `build.rs`.
See https://github.com/rust-lang/cargo/issues/4789 which also offers the solution adopted here.

### DIFF
--- a/butane/Cargo.toml
+++ b/butane/Cargo.toml
@@ -28,7 +28,8 @@ log = ["butane_core/log"]
 r2d2 = ["dep:r2d2"]
 tls = ["butane_core/tls"]
 uuid = ["butane_codegen/uuid", "butane_core/uuid"]
-# This feature is for testing only. It will delete the .butane directory. You probably don't want that.
+# This feature is for testing only. It will delete the .butane directory inside the butane crate, which only
+# exists when running tests.  It has no effect when running butane as a dependency.
 _auto_delete_dot_butane = []
 
 [dependencies]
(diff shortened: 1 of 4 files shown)


butane/Cargo.toml
```diff
@@ -28,6 +28,8 @@ log = ["butane_core/log"]
 r2d2 = ["dep:r2d2"]
 tls = ["butane_core/tls"]
 uuid = ["butane_codegen/uuid", "butane_core/uuid"]
+# This feature is for testing only. It will delete the .butane directory. You probably don't want that.
+_auto_delete_dot_butane = []
```
Electron100, line 32: Will this be an issue when running with --all-features?

jayvdb: `--all-features` will include this, yes, but it is harmless to run these bits of the `build.rs`, both
- for us, as `cargo build --all-features` will unnecessarily delete `.butane`, and
- for end-users as the butane crate shouldnt contain a `.butane` subdirectory.

Probably I should reword this note to be a little less scary.

butane/build.rs
```diff
@@ -1,16 +1,22 @@
...
+    println!("cargo:rerun-if-changed=tests/");
+
+    // Note that #[cfg(test)] and CARGO_CFG_TEST do not work in build.rs
+    // See https://github.com/rust-lang/cargo/issues/4789
+    #[cfg(feature = "_auto_delete_dot_butane")]
+    {
+        clean_dot_butane();
     }
-    println!("cargo:rerun-if-changed=tests");
+}
```
jayvdb, line 10: We could also reduce how often this runs by disabling it when the `PROFILE` = `release`
https://stackoverflow.com/questions/59277384/is-it-possible-to-invoke-a-build-rs-only-for-release-mode

The env will be quicker to check than looking for existence of a directory on the filesystem.

Electron100 (approved):

PR #354, opened 2025-05-17
Disable logging of macro expansion

jayvdb:
`cargo ... --all-features` is very noisy because of `maybe-async-cfg/debug`.  I think it has very little utility for end-users; we can enabled it locally when debugging.

It was added in https://github.com/Electron100/butane/commit/b2d7426de7ca5b0169f7cc33728c480681cc7131


Electron100:
Should we just remove the debug feature since it's now just an alias for log?

Electron100 (approved):

jayvdb:
The debug feature goes back a long way https://github.com/Electron100/butane/commit/e3cf0d35a85cf0fa74a57b62681872ddf6ff27d3

Also we have some code which is activated by the feature
https://github.com/Electron100/butane/blob/76d652676374bf958dfe29606488b7bbef34b32d/butane_core/src/db/pg.rs#L75

We could also only add `Debug` impls when the debug feature is present; I've seen that done in other crates.

PR #355, opened 2025-05-17
Allow any valid sqlite or pg connection string

jayvdb:
Continuation of https://github.com/Electron100/butane/pull/344 to address the `BUTANE_PG_CONNSTR` part of https://github.com/Electron100/butane/issues/343 , and also ensure that `butane init` does not need the first parameter to define the "backend name".

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,7 @@ jobs:
         sudo cat /etc/postgresql/16/main/pg_hba.conf
         sudo service postgresql restart && sleep 3
         echo BUTANE_PG_CONNSTR="host=localhost user=postgres sslmode=disable port=5432" >> $GITHUB_ENV
+        echo "/usr/lib/postgresql/16/bin" >> $GITHUB_PATH
     - name: Setup PostgreSQL on MacOS
       if: runner.os == 'macOS'
       run: |
@@ -109,6 +110,8 @@ jobs:
       run: |
         set -ex
         make regenerate-example-migrations
+        # TODO: This file is created by one of the tests on Windows
+        rm -f butane_core/sqlite
         git add -A
         git diff --cached --exit-code
     - name: Run tests in examples
--- a/butane_core/src/db/mod.rs
+++ b/butane_core/src/db/mod.rs
@@ -526,6 +526,9 @@ pub trait Backend: Send + Sync + DynClone {
 
 dyn_clone::clone_trait_object!(Backend);
 
+/// Regular expression to match PostgreSQL key-value pairs in a connection string.
+const PG_KEY_PAIR_RE: &str = r"(hostaddr|host|dbname|user|port)\s*=";
+
 /// Connection specification. Contains the name of a database backend
 /// and the backend-specific connection string. See [`connect`]
 /// to make a [`Connection`] from a `ConnectionSpec`.
@@ -549,7 +552,7 @@ impl ConnectionSpec {
         contents.push('\n');
         f.write_all(contents.as_bytes()).map_err(|e| e.into())
     }
-    /// Load a previously saved connection spec
+    /// Load a previously saved connection spec.
     pub fn load(path: impl AsRef<Path>) -> Result<Self> {
         let path = conn_complete_if_dir(path.as_ref());
         serde_json::from_reader(fs::File::open(path)?).map_err(|e| e.into())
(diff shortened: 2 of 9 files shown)


butane_core/tests/connection.rs
```diff
@@ -7,30 +9,366 @@ async fn connection_not_closed(conn: ConnectionAsync) {
     assert!(!conn.is_closed());
 }
 
+// The SQLite connection URI tests cover most cases described at https://www.sqlite.org/c3ref/open.html
+// and https://www.sqlite.org/inmemorydb.html
+
+// TODO: verify the absolute tests on Windows use a drive letter prefix.
+// (some of the absolute tests may not pass on Windows)
```
jayvdb, line 16: TODO as part of this PR

butane_core/src/db/mod.rs
```diff
@@ -560,28 +560,116 @@ impl ConnectionSpec {
...
         }
     }
+    /// Get the backend name.
+    pub const fn backend_name(&self) -> &String {
+        &self.backend_name
+    }
+    /// Get the database connection string.
+    pub const fn connection_string(&self) -> &String {
+        &self.conn_str
+    }
```
jayvdb, line 570: Medium term objective is to remove the `pub` on these two fields, but this PR is big enough.

.github/workflows/ci.yml
```diff
@@ -109,6 +110,8 @@ jobs:
       run: |
         set -ex
         make regenerate-example-migrations
+        # TODO: This file is created by one of the tests on Windows
+        rm -f butane_core/sqlite
```
jayvdb, line 114: I'll solve this on a different branch afterwards - need to figure out which one(s) it is, and delete the file as part of those tests.

butane_core/src/db/mod.rs
```diff
@@ -560,28 +560,116 @@ impl ConnectionSpec {
...
+    /// Get the connection string URI if it is a URI.
+    /// Returns None if the connection string is not a URI.
+    pub fn connection_string_uri(&self) -> Option<url::Url> {
+        url::Url::parse(self.connection_string()).ok()
+    }
+
+    fn is_pg_key_value_pairs(connection_string: &str) -> bool {
+        // host and hostaddr are optional in Postgresql, however this prevents connecting
+        // without one of them: https://github.com/sfackler/rust-postgres/issues/1239
+        connection_string.contains('=')
```
Electron100, line 580: Shouldn't we check for the backend type being pg too? Theoretically another backend could use similar syntax

butane_core/src/db/mod.rs
```diff
@@ -560,28 +560,116 @@ impl ConnectionSpec {
...
+        &self.conn_str
+    }
+    /// Get the connection string URI if it is a URI.
+    /// Returns None if the connection string is not a URI.
+    pub fn connection_string_uri(&self) -> Option<url::Url> {
+        url::Url::parse(self.connection_string()).ok()
+    }
+
+    fn is_pg_key_value_pairs(connection_string: &str) -> bool {
+        // host and hostaddr are optional in Postgresql, however this prevents connecting
```
Electron100, line 578: I'm not sure if I'm quite following how this comment matches the code below. It looks like you're returning true if any of the known parameters are found, not just host or hostaddr

butane_core/src/db/mod.rs
Electron100, line 580: What's the point of checking for "=" at line 580, since the options you && with all have = also?

Electron100: Overall the parsing here seems a bit fragile. Why not see if `tokio_postgres::config::Config::from_str` succeeds (and also check if it fails to parse as a URL if necessary)

butane_core/src/db/mod.rs
```diff
@@ -560,28 +560,116 @@ impl ConnectionSpec {
...
+    }
+
+    /// Add a query parameter to the connection string.
+    pub fn add_parameter(&mut self, name: &str, value: &str) {
+        if Self::is_pg_key_value_pairs(&self.conn_str) {
+            // Append using PostgreSQL key-value pair syntax.
+            self.conn_str.push_str(&format!(" {}={}", name, value));
+            return;
+        }
+        if self.conn_str.contains('?') {
```
Electron100, line 600: I think we should check if the connection string is a URL first. If it's not a URL and it's not postgres key-value pairs then we don't really know how to add a parameter and should return an error

butane_test_helper/Cargo.toml
```diff
@@ -12,6 +12,7 @@ repository.workspace = true
 documentation = "https://docs.rs/butane/"
 
 [features]
+bin = ["ctrlc"]
```
Electron100, line 15: Unless there's a tight connection I'm missing, could you put the butane_test_helper changes in a separate PR? They seem logically distinct.

butane_core/src/db/mod.rs
jayvdb: This function doesnt use self as it is also used by the `TryFrom`

jayvdb: I'll take a look at that function upstream, however I am keen to follow the postgresql server rules, not tokio-postgres rules which ive found to be rather buggy.

The first `contains('=')` here is a short circuit to avoid all of the other checks which will be more expensive.  The algorithm can be improved, but works correctly atm.

butane_core/src/db/mod.rs
jayvdb: I'll check that, but IIRC SQLite supported `?` for things that were not a valid URI

butane_test_helper/Cargo.toml
jayvdb: I can move it to a separate PR, but I have been developing them in concert because the test helper is needed to test the connection strings

butane_core/src/db/mod.rs
jayvdb: And it is not `pub`.

And the design of this PR is intended to allow only existing valid sqlite & pg "conn_str" values to be supported.
In the future IMO butane should require any new backend is implemented so that connection strings are a URI which contain a valid scheme, even if it is pseudo scheme like `sqlite:...` implemented in this PR.

and we might in the future deprecate some of the stranger sqlite & pg connection strings like `:memory:` (which `url::Url::parse` rejects, and sqlite interprets strangely when params are added) , recommending that people use the equivalent `file::memory:` / `sqlite::memory:`.

Electron100 (approved):

Issue #353, opened 2025-05-16
Many.add and .remove require reloading the Many before .get will work


jayvdb:
This has caught me out a few times, as it breaks from what the developer would expect.

```rust
let _ = post.tags.load(&conn).unwrap();

// do other stuff

post.tags.add(&tag_european).unwrap();
// the above .add causes .get() now to fail, but that wouldnt be so bad, as .add() would normally be followed by .save()
post.save(&conn).await.unwrap();
let _ = post.tags.get().unwrap(); // Boom. ValueNotLoaded
```

The `Many.add` and `Many.remove` both do the following for good reason 
```rust
// all_values is now out of date, so clear it
self.all_values = OnceLock::new();
```

However if the Many relationship has a lot of rows, doing a `.load()` when only a single item has been added or removed is undesirable.

One approach would be to add `Many.add_without_clearing(&mut self, val: T)`, so that `val` can be put included in `.all_values`.  "_without_clearing" isnt good, but is the best I can come up with quickly.  A `delete_without_clearing` would also delete from `.all_values`.

I would not mind if `.get()` becomes unusable after a `.add`/`.remove` if it became usable again when `.save` has been completed.

This could be achieved by maintaining a separate `.already_loaded_values` whenever a .add or .remove, and then a `.save_and_keep_local_changes()` could set `.all_values = .already_loaded_values + fresh_fetch_of_added - removed` .

Any other alternatives?

RobertoMaurizzi:
Django hasL
-  .set() to set a list of FK/instances, 
- .add() a list
- .remove() a list
- .clear() removes all

jayvdb:
Thanks for suggesting we look at Django for inspiration.

https://docs.djangoproject.com/en/5.1/ref/models/relations/ is the best starting point for understanding Django 's implementation of `Many`.  Most important is this bit

> Note that add(), aadd(), create(), acreate(), remove(), aremove(), clear(), aclear(), set(), and aset() all apply database changes immediately for all types of related fields. In other words, there is no need to call save()/asave() on either end of the relationship.

Applying changes immediately is IMO much better wrt relationships.

Rather than worrying about the existing .add and .remove, IMO 
`.set()` and `.clear()` would be wonderful additions, as they dont conflict or confusion with the existing method names, and we can implement them like Django, i.e. they perform a transition in the DB and in the ORM immediately, so `.get()` can be invoked afterwards without an error.

`.clear()` is just `.set([])`.

[...]

PR #357, opened 2025-05-17
Add Many.set

jayvdb:
Suggestion raised at https://github.com/Electron100/butane/issues/353

### DIFF
--- a/butane_core/src/many.rs
+++ b/butane_core/src/many.rs
@@ -198,16 +198,26 @@ where
     async(feature = "async")
 )]
 pub trait ManyOps<T: DataObject> {
+    /// Save all unsaved relation changes to the backend.
+    ///
     /// Used by macro-generated code. You do not need to call this directly.
+    ///
+    /// This will insert added values first, and then remove the removed values.
+    /// Use inside a transaction to provide atomicity.
     async fn save(&mut self, conn: &impl ConnectionMethods) -> Result<()>;
 
     /// Delete all references from the backend, and any unsaved additions.
+    ///
+    /// This operation is atomic.
     async fn delete(&mut self, conn: &impl ConnectionMethods) -> Result<()>;
 
     /// Overwrite the references in the backend, and clears unsaved additions.
+    ///
+    /// This will call `delete()` first, and then insert the items scheduled to be added.
+    /// Use inside a transaction to provide atomicity.
     async fn set(&mut self, conn: &impl ConnectionMethods, values: Vec<T>) -> Result<()>;
 
-    /// Loads the values referred to by this many relationship from the backend.
+    /// Loads the values referred to by this many relationship from the backend if necessary.
     async fn load<'a>(
         &'a self,
         conn: &impl ConnectionMethods,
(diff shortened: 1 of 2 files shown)


butane_core/src/many.rs
```diff
@@ -195,20 +195,21 @@ pub trait ManyOps<T: DataObject> {
...
     async fn load<'a>(
         &'a self,
         conn: &impl ConnectionMethods,
     ) -> Result<impl Iterator<Item = &'a T>>
     where
         T: 'a;
 
-    /// Loads and orders the values referred to by this many relationship from a
-    /// database if necessary and returns a reference to them.
+    /// Loads and orders the values referred to by this many relationship from the backend.
```
jayvdb, line 212: I believe the "if necessary" is misleading..?

And "returns a reference to them" seems self-evident from the function signature, and isnt surprising or important if the end-user isnt made aware of this.

butane_core/src/many.rs
```diff
@@ -263,8 +264,20 @@ impl<T: DataObject> ManyOps<T> for Many<T> {
...
+        Ok(())
+    }
+
+    async fn set(&mut self, conn: &impl ConnectionMethods, values: Vec<T>) -> Result<()> {
+        ManyOps::delete(self, conn).await?;
+
+        for value in &values {
+            self.add(value)?;
+        }
+        ManyOps::save(self, conn).await?;
```
jayvdb, line 278: The perf of this can be vastly improved using bulk inserts, but IMO that can be done later, after the API of this new method has been finalised.

Also will probably require new DML abstractions for the bulk inserts.

If the Many object is already loaded, doing a set difference could also help avoiding performing unnecessary deletions and insertions, but I worry about that because the Many internals may be out of sync with the database if they have been modified by another connection to the database.

butane_core/src/many.rs
Electron100: It goes through a OnceLock, so IIRC it does not load if already loaded.

Agree about the return part being redundant

butane_core/src/many.rs
jayvdb: Created https://github.com/Electron100/butane/issues/359 for the DML abstraction for bulk inserts.

butane_core/src/many.rs
```diff
@@ -195,20 +195,21 @@ pub trait ManyOps<T: DataObject> {
     /// Used by macro-generated code. You do not need to call this directly.
     async fn save(&mut self, conn: &impl ConnectionMethods) -> Result<()>;
 
-    /// Delete all references from the database, and any unsaved additions.
+    /// Delete all references from the backend, and any unsaved additions.
     async fn delete(&mut self, conn: &impl ConnectionMethods) -> Result<()>;
 
-    /// Loads the values referred to by this many relationship from the
-    /// database if necessary and returns a reference to them.
+    /// Overwrite the references in the backend, and clears unsaved additions.
```
Electron100, line 201: The comment should probably caution that this is done non-atomically (the deletion of the previous values may be observed before the new values are set). Alternatively require a `Connection` so that we can actually do it in a transaction

Electron100: I realize this is true of the way `save` is implemented today in general, but the fact that this method may replace with an overlapping set may make the fact that items present in both the old and the new may be briefly missing surprising.

Electron100 (approved):

butane_core/src/many.rs
jayvdb: I would prefer that the user is responsible for transactions.

Updated save/delete/set to indicate whether they are atomic, and recommend using a transaction if atomic is desired.

Issue #360, opened 2025-05-21
`r#` identifiers not working as column names


Nekidev:
I was trying to name one of my table's columns `type`, but `r#type` causes the macro to panic.

```rs
proc-macro panicked: `"r#type"` is not a valid identifier
```

A minimal reproduction of the error can be created with

```rs
#[model]
#[table = "table_name"]
#[derive(Debug)]
pub struct MyTable {
    #[pk]
    id: String,

    r#type: MyTableType  // Enum defined with #[derive(FieldType)]
}
```

jayvdb:
https://github.com/Electron100/butane/pull/364 should fix this problem.  I am going to keep going, as I suspect there are other places where `r#` can be used that might break butane.  Might as well find and fix them all now.

PR #364, opened 2025-06-01
Support r# in struct and member names

jayvdb:
Closes https://github.com/Electron100/butane/issues/360


butane/tests/json.rs
```diff
@@ -278,3 +278,31 @@ async fn inline_json(conn: ConnectionAsync) {
...
+        r#type: r#type,
+    }
+
+    let mut foo = StructWithReservedWordMember::default();
+    foo.id = "1".to_string();
+    let expected = r#type {
+        r#type: "test".to_string(),
+    };
+    foo.r#type = expected.clone();
+    foo.save(&conn).await.unwrap();
```
jayvdb, line 304: This unwrap on line 304 fails when `struct r#type { ... }` is not `pub`.
```
thread 'r_hash_field_type_newtype_json_async_pg' panicked at butane/tests/json.rs:304:27:
called `Result::unwrap()` on an `Err` value: Postgres(Error { kind: ToSql(1), cause: Some(WrongType { postgres: Text, rust: "serde_json::value::Value" }) })
```

jayvdb: ah, that is part of it, but the cause is that butane/ now has a model called `r#type` and a field type called `r#type`

butane/tests/basic.rs
```diff
@@ -514,3 +514,43 @@ struct TypeVariantsTest {
...
+    pub struct StructWithReservedWordMember {
+        id: String,
+
+        r#type: String,
+    }
+
+    let mut foo = StructWithReservedWordMember::default();
+    foo.id = "1".to_string();
+    foo.r#type = "test".to_string();
+    foo.save(&conn).await.unwrap();
```
jayvdb, line 532: now this fails on line 532 with
```
thread 'r_hash_struct_member_async_pg' panicked at butane/tests/basic.rs:532:27:
called `Result::unwrap()` on an `Err` value: Postgres(Error { kind: ToSql(1), cause: Some(WrongType { postgres: Jsonb, rust: "&str" }) })
```

PR #369, opened 2025-06-06
Ensure all syn::Type are a syn::Path without qself

jayvdb:
See https://docs.rs/syn/latest/syn/struct.TypePath.html which defines a `qself`, and I believe that use with a `qself` would break butane, and probably couldnt be used in the vast majority of the places where butane ForiegnKey & Many relationships appear.

### DIFF
--- a/butane_core/src/codegen/mod.rs
+++ b/butane_core/src/codegen/mod.rs
@@ -1,4 +1,9 @@
-//! Code-generation backend
+//! Code-generation backend.
+//!
+//! Any public member is only for use in `butane_codegen` crate, and
+//! all members of this module are hidden from the documentation, and
+//! may change at any time without warning.
+#![doc(hidden)]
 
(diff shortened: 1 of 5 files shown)


butane_core/src/codegen/mod.rs
```diff
@@ -280,45 +285,39 @@ fn fields(ast_struct: &ItemStruct) -> impl Iterator<Item = &Field> {
...
+    if let syn::Type::Path(type_path) = ty {
+        if type_path.qself.is_some() {
+            panic!("TypePath with qself is not supported by butane");
         }
-        .into();
+        return &type_path.path;
+    } else {
+        panic!("butane codegen only supports TypePath types");
+    }
+}
```
jayvdb, line 300: Happy to return a `Result` if there is any chance that a `qself` could be valid in the definition of a model, field-type, etc that butane comes into context with.

Electron100 (approved):

PR #386, opened 2025-10-25
Make write of types.json atomic

jayvdb:
This is what Claude Sonnet 4.5 believes is the problem with https://github.com/Electron100/butane/issues/162


jayvdb:
Possibly this error is now the same problem appearing in a new way
https://github.com/Electron100/butane/actions/runs/18809595881/job/53668741071?pr=387

jayvdb:
Ya I am not convinced this is a fix, and the CI runs demonstrate it only changed the errors.  I'd prefer to keep the same errors until we really find the cause.

PR #390, opened 2025-10-28
CLI: squash

jayvdb:
Semi-related to https://github.com/Electron100/butane/issues/91

A common case is a migration has been committed on a PR, but needs amending before approval & merging.

### DIFF
--- a/butane_cli/src/lib.rs
+++ b/butane_cli/src/lib.rs
@@ -635,6 +635,66 @@ pub fn collapse_migrations(base_dir: &PathBuf, new_initial_name: Option<&String>
     Ok(())
 }
 
+/// Squash the most recent migration into the previous migration.
+pub fn squash_migration(base_dir: &Path) -> Result<()> {
+    let mut ms = get_migrations(base_dir)?;
+
+    let all_migrations = ms.all_migrations().unwrap_or_else(|e| {
+        eprintln!("Error: {e}");
+        std::process::exit(1);
+    });
+
+    if all_migrations.len() < 2 {
+        eprintln!("Cannot squash migration: need at least 2 migrations");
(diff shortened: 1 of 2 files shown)


butane_cli/src/lib.rs
```diff
@@ -635,6 +635,66 @@ pub fn collapse_migrations(base_dir: &PathBuf, new_initial_name: Option<&String>
...
+    // Get the database state from the latest migration (which includes all changes)
+    let target_db = latest_migration.db()?;
+
+    // Remove the two old migrations from the state first
+    // Note: We need to be careful about the order here
+    ms.detach_latest_migration()?; // Remove latest
+    ms.detach_latest_migration()?; // Remove second-to-last
+
+    // Create a new migration that represents the squashed state, using the name of the second-to-last migration
+    ms.create_migration_to(&backends, &second_to_last_name, from_migration, target_db)?;
```
jayvdb: This overwrites the .sql in the previous migration.

Maybe we need a prompt to ensure users understand this.

butane_cli/src/main.rs
```diff
@@ -71,6 +71,8 @@ However if the migration has been manually edited, it will need to be manually r
         /// Name to use for the new migration.
         name: String,
     },
+    /// Merge the most recent migration into the previous migration.
+    Squash,
```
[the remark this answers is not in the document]
Electron100: I think I have a slight preference for SquashMigration

butane_cli/src/lib.rs
Electron100: Yeah, I think this is the right behavior, but probably good to introduce user confirmation for (ideally with a flag to override the prompt if necessary to use it in a fully automated fashion)

PR #393, opened 2025-10-30
Add mysql backend

jayvdb:
Vibe coded - needs cleanup

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,10 +25,10 @@ jobs:
     strategy:
       matrix:
         include:
-          #- os: ubuntu-latest
-          #  mise_tools: "ephemeral-postgres mysql"
-          #- os: macos-13
-          #  mise_tools: "ephemeral-postgres mysql"
+          - os: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -42,17 +45,18 @@ jobs:
     - uses: actions/checkout@v4
       # Postgresql setup adapted from Diesel CI
       # Disable ssl as server doesn't have a trusted cert
-    - name: Setup PostgreSQL on Linux
+    - name: Setup PostgreSQL & libaio1t64 on Linux
```
jayvdb: should be just libaio

.github/workflows/ci.yml
```diff
@@ -42,17 +45,18 @@ jobs:
...
       if: runner.os == 'Linux'
       run: |
         set -ex
         sudo apt-get update
-        sudo apt-get install -y postgresql
+        sudo apt-get install -y libaio-dev postgresql
```
jayvdb: installing libaio1t64 should be sufficient

PR #396, opened 2025-10-31
Add support for fixed length strings

### LINKED ISSUE #392: Fixed length strings
IMO this is a fairly important missing feature.

Some potential solutions
- https://docs.rs/fixed-size/latest/fixed_size/
- https://docs.rs/fixed_str/latest/fixed_str/
- https://docs.rs/arrayvec/latest/arrayvec/struct.ArrayString.html - not sure this will work
- https://crates.io/crates/deku_string
- https://crates.io/crates/microstr
- https://docs.rs/heapless/0.9.1/heapless/string/type.String.html

jayvdb:
Closes #392 

I am not advocating `arrayvec::ArrayString` - it looks wrong, because dbs store chars, but this crate is about bytes.


butane_core/Cargo.toml
```diff
@@ -24,6 +24,7 @@ tls = ["native-tls", "postgres-native-tls"]
 turso = ["async", "dep:turso"]
 
 [dependencies]
+arrayvec = "0.7"
```
Electron100: I'd prefer for arrayvec (or whichever we land on here) to be an optional dependency

Electron100 (commented):
I don't think I've quite understood what you were trying to achieve here. From #392 I initially thought you were going for something that would be represented as a fixed-length string through to the SQL backend (e.g. CHAR(n) in Postgres). This is just supporting ArrayStr as a usable type in models that ultimately goes through to TEXT. Which is OK, but I want to make sure I've understood your intention correctly?

jayvdb:
The goal is to be using VARCHAR(x) in the schema def.  So far, this is just getting the Rust side of things.  Next step is Switching from arraayvec to a type that can match VARCHAR(x) of utf8.  Then fix the migrations to use VARCHAR(x).

PR #397, opened 2025-10-31
Fix pg constraint drops

jayvdb:
We dont explicitly name the constraints at https://github.com/Electron100/butane/blob/master/butane_core/src/db/pg.rs#L681-L687

And it appears that pg auto naming of constraints uses the exact casing of the table.


jayvdb:
I also want to add tests for this.

PR #404, opened 2025-11-01
Add libsql support

### LINKED ISSUE #395: libsql-server support
We have basic turso support now; c.f. https://github.com/Electron100/butane/issues/287

turso also has TCPIP support

https://mise.jdx.dev/registry.html includes `libsql-server` so we can use that to set up the testing.

This could be as easy as adding libsql server setup in butane_test_helper , and improving `butane_core/tests/connection.rs` to include libsql uri tests.

jayvdb:
Closes #395


butane_core/src/db/sqlite.rs
```diff
@@ -885,7 +905,12 @@ pub fn sql_insert_or_update(table: &str, columns: &[Column], pkcol: &Column, w:
         ", "
     });
     write!(w, ")").unwrap();
-    write!(w, " ON CONFLICT ({}) DO ", pkcol.name()).unwrap();
+    write!(
+        w,
+        " ON CONFLICT ({}) DO ",
+        helper::quote_reserved_word(pkcol.name())
+    )
+    .unwrap();
```
jayvdb: this fixes a bug in the sqlite impl

butane_test_helper/src/libsql.rs
```diff
@@ -0,0 +1,302 @@
...
+                    return Ok(path.to_string());
+                }
+            }
+        }
+
+        // If not found, try to install it
+        log::warn!("sqld not found, attempting to install via cargo...");
+        let install_output = std::process::Command::new("cargo")
+            .args(["install", "libsql-server", "--bin", "sqld"])
+            .output()?;
```
jayvdb: not sure about this bit

Issue #226, opened 2024-03-26
rollback fails on sqlite when data exists


jayvdb:
For the test being added in https://github.com/Electron100/butane/pull/223 , the sqlite backend fails if data is inserted.

```
thread 'migrate_and_rollback_sqlite' panicked at examples/getting_started/tests/rollback.rs:49:67:
rollback of 20240115_023841384_dbconstraints failed: Sqlite error FOREIGN KEY constraint failed
```

PR #407, opened 2025-11-02
Disable defer_foreign_keys during sqlite migration

jayvdb:
Fixes one of https://github.com/Electron100/butane/issues/226


butane_core/src/db/sqlite.rs
```diff
@@ -858,7 +858,9 @@ fn change_column(
         Some(col) => new_table.replace_column(col.clone()),
         None => new_table.remove_column(old.name()),
     }
-    let stmts: [&str; 4] = [
+    // Wrap in PRAGMA defer_foreign_keys to allow dropping tables that are referenced
+    let stmts: [&str; 6] = [
+        "PRAGMA defer_foreign_keys = ON;",
```
jayvdb: not sure I like this, but it does solve one of our rollback problems.

PR #409, opened 2025-11-02
Use expectorate to snapshot generated SQL

jayvdb:
Uses https://crates.io/crates/expectorate

Turns on turso backend for the adb and migration tests.


butane_core/tests/migration.rs
```diff
@@ -324,45 +190,83 @@ fn test_migrate(
...
+    // For turso, certain migrations fail due to turso-specific bugs with ALTER TABLE
+    // For tests that add/remove fields, turso panics with "table being renamed should be in schema"
+    if backend_name == "turso" && test_name.starts_with("add") {
+        let result = ms.migrate_async(conn).await;
+        assert!(
+            result.is_err(),
+            "Expected turso migration to fail with 'table being renamed should be in schema' error"
+        );
+        return;
+    }
```
jayvdb: just noting there are 3 or 4 tests here which match prefix "add"

butane_core/tests/migration.rs
```diff
@@ -543,19 +472,42 @@ fn migration_delete_table(conn: &mut Connection, expected_up_sql: &str, expected
...
 
     // Now downgrade, just to make sure we can
-    ms.unmigrate(conn).unwrap();
-
-    let to_apply = ms.unapplied_migrations(conn).unwrap();
-    assert_eq!(to_apply.len(), 2);
+    // Skip unmigrate for sqlite and turso delete_table test due to known issue
+    if backend_name == "sqlite" || backend_name == "turso" {
+        return;
+    }
```
jayvdb: again, this case was only being run on pg previously.

.github/workflows/ci.yml
```diff
@@ -96,6 +96,8 @@ jobs:
       with:
         tool: cargo-deny,editorconfig-checker,mise,typos
     - name: Install ephemeral-postgres via mise
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
Electron100: How is this  related to expectorate?

Electron100: Would like to understand this before merge, but no other concerns with the change

jayvdb:
It would be good to get this in so I can use snapshots in the other PRs

PR #411, opened 2025-11-03
Use mem migrations to build models on WASM

jayvdb:
Sits on top of https://github.com/Electron100/butane/pull/410

A bit more controversial, but IMO worth the pain.

Notably this doesnt allow models to be referenced before they appear in the source, hence the change to getting_started models order.  And thus no circular models.  We can document these as part of the many limitations for the WASM target.


jayvdb:
fwiw, I know what the problem is with the newtype example, but as is often the case, it is going to require enhancing the migration generator, specifically supporting converting a primary column from pg bytea to jsonb, whilst there are foreign key references to it.  I'm half way there ..

Issue #412, opened 2025-11-06
Non-unique binary names causing issues


jayvdb:
I believe this is contributing to CI failures.  They are intermitted, only on Windows.  Probably both getting compiled and linked at the same time causing occasional clashes.

https://github.com/Electron100/butane/actions/runs/19120967749/job/54641222303?pr=410

CI logs show
```
warning: output filename collision.
The bin target `publish_post_async` in package `getting_started_async v0.1.0 (D:\a\butane\butane\examples\getting_started_async)` has the same output filename as the bin target `publish_post_async` in package `getting_started v0.1.0 (D:\a\butane\butane\examples\getting_started)`.
Colliding filename is: D:\a\butane\butane\target\debug\publish_post_async.pdb
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
   Compiling turso_core v0.3.0
   Compiling butane v0.8.1 (D:\a\butane\butane\butane)
   Compiling example v0.1.0 (D:\a\butane\butane\example)
   Compiling turso v0.3.0
   Compiling butane_core v0.8.1 (D:\a\butane\butane\butane_core)
   Compiling butane_test_helper v0.8.1 (D:\a\butane\butane\butane_test_helper)
   Compiling getting_started v0.1.0 (D:\a\butane\butane\examples\getting_started)
   Compiling getting_started_async v0.1.0 (D:\a\butane\butane\examples\getting_started_async)
   Compiling butane_cli v0.8.1 (D:\a\butane\butane\butane_cli)
   Compiling custom_pg v0.1.0 (D:\a\butane\butane\examples\custom_pg)
   Compiling newtype v0.1.0 (D:\a\butane\butane\examples\newtype)
   Compiling reserved-words v0.1.0 (D:\a\butane\butane\examples\reserved-words)
error: linking with `link.exe` failed: exit code: 1104
  |
[...]

PR #413, opened 2025-11-06
Fix example binary name

jayvdb:
Fixes https://github.com/Electron100/butane/issues/412


jayvdb:
Trivial - just a mistake I made recently.

